PR #1, opened 2025-08-29
tests: SUI-712  Initial System test

StuwiiDev:
All steps fail until we start building out the system.

Also updated gitignore and added sln file for initial project build


SingleUniqueIdentifier.sln
```diff
@@ -0,0 +1,29 @@
...
+	GlobalSection(NestedProjects) = preSolution
+		{B6B6C645-B718-47CC-90EC-EE703DC6905B} = {101D4708-21CF-4789-8628-D37580CC6927}
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{B6B6C645-B718-47CC-90EC-EE703DC6905B}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{B6B6C645-B718-47CC-90EC-EE703DC6905B}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{B6B6C645-B718-47CC-90EC-EE703DC6905B}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{B6B6C645-B718-47CC-90EC-EE703DC6905B}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+EndGlobal
```
jtaylor100, line 29: Wondering whether it's a good idea to have a single solution for the whole repo or not — kind of ties all of the work streams together in a dependency.

Probably my preferred structure would be a solution for the 'find' and 'access' workstreams — and one more for system-level tests which would contain the 'System.E2ETest' project. What do you think?

jtaylor100 (approved):

PR #562, opened 2026-04-08
added ability to renew lease

tomhawkin:


Apps/Find/src/SUI.Find.Infrastructure/Services/JobClaimService.cs
```diff
@@ -91,6 +91,59 @@ CancellationToken cancellationToken
...
+        string jobId,
+        string leaseId,
+        CancellationToken cancellationToken
+    )
+    {
+        var job = await GetCurrentLeasedJobForCustodian(custodianId, jobId, cancellationToken);
+        var utcNow = timeProvider.GetUtcNow();
+
+        if (
+            job == null
```
robshakespeare: **nitpick:** I think its worth including an explicit check to verify that the job is leased:


```suggestion
            job == null
            || job.LeaseId == null
```

Apps/Find/src/SUI.Find.Infrastructure/Services/JobClaimService.cs
```diff
@@ -91,6 +91,59 @@ CancellationToken cancellationToken
...
+            || job.LeaseId != leaseId
+            || job.LeaseExpiresAtUtc == null
+            || job.LeaseExpiresAtUtc <= utcNow
+        )
+        {
+            return null;
```
robshakespeare, line 126: **suggestion:** Please could you update this to log a warning, and include `job?.LeaseId` and `job?.LeaseExpiresAtUtc` in the log message

robshakespeare (commented):
Looks good, just a couple of small suggestions

Apps/Find/src/SUI.Find.Infrastructure/Services/JobClaimService.cs
robshakespeare: Nice, thanks for updating this

robshakespeare (approved):
Nice, looks great

PR #563, opened 2026-04-13
ADRs updates

robshakespeare:


robshakespeare:
This has mostly become obsolete or totally changed, now that the scope has changed.

It feels like the best decision is to close this off, and revisit in 6-12 months when the next phase becomes established.

PR #564, opened 2026-04-13
chore: Job.SearchingOrganisationId should be optional, it is only required in the context of Search jobs

robshakespeare:


udara-jay (approved):
Looks good 👍

PR #565, opened 2026-04-14
fix: SUI-1668 - use valid storage property names when marking job as completed

robshakespeare:
Use valid Azure Table Storage property names when marking job as completed, fixes `PropertyNameInvalid` error in deployed environments (SUI-1668).


udara-jay (approved):
Looks good to me 👍

PR #566, opened 2026-04-14
Test: SUI-1611 - E2E Test: New Organisation that has no data sharing agreements

udara-jay:


Apps/StubCustodians/src/SUI.StubCustodians.API/Data/faux-01.custodian.json
```diff
@@ -0,0 +1,6 @@
+{
+  "orgId": "FAUX-01",
+  "orgName": "Example Third Party Organisation",
```
robshakespeare: **nitpick:** Given the discovery of "other" already being used, and hence us switching to use the term "faux" for this, please could you update the Org Name (in both files) to be:

```suggestion
  "orgName": "Faux Organisation for Negative Testing",
```

robshakespeare (dismissed):
Looks great, thank you for making these updates and persevering to get this test case working.

Just one nitpick change please, that has come out of the other -> faux change of direction.  Many thanks.

robshakespeare (approved):
Nice one, thank you

PR #567, opened 2026-04-14
feat: SUI-1671 - improve e2e test reliability immediately after deployment

robshakespeare:
Updates the e2e tests to wait for the latest version to be deployed and started before the tests commence.  Rather than wait for an arbitrary period of time (which could wait for too long unnecessarily), it waits until the API responds with a build timestamp which is on or after a known point in time that confirms the latest version is in use.

#### Context and background info:
The e2e tests falsely fail when they are automatically ran after deployment.  They fail because the e2e tests run immediately after deployment, at which point the deployed API reports that is is up and healthy because the new deployment hasn’t started up yet.  Once the function app restarts, the API then temporarily times-out.  After a non-determinate period of time, the new version of the API is up and healthy.  The changes in this PR ensure that the e2e tests wait for the new version to be deployed and up and running.


udara-jay (approved):
Looks great! thanks for doing this! 👍

robshakespeare:
> Looks great! thanks for doing this! 👍

Thank you :+1:

PR #571, opened 2026-04-15
feat: SUI-1613 - Include auditing after PEP filtering in the Polling Architecture

jordanstarksbrowning-hippo:


Apps/Find/src/SUI.Find.Application/Models/JobContext.cs
```diff
@@ -0,0 +1,7 @@
+namespace SUI.Find.Application.Models;
+
+public sealed record JobContext(
```
robshakespeare: **nitpick:** This is just a tuple used internally with the `JobResultHandler`, it makes sense to not expose this type.  Please can you change it back to being a private class with the `JobResultHandler`?  Maybe it was moved out during refactoring in this ticket?  But it has ended up still being only used within the `JobResultHandler` class.

Apps/Find/src/SUI.Find.Application/Interfaces/IPolicyEnforcementService.cs
```diff
@@ -0,0 +1,29 @@
...
+
+public interface IPolicyEnforcementAndAuditingService
+{
+    Task<IReadOnlyList<PepResultItem<TItem>>> FilterItemsAndAuditAsync<TItem>(
+        PepFilterAndAuditInput<TItem> input,
+        CancellationToken cancellationToken
+    )
+        where TItem : IPepFilterable;
+
+    Task CreateAndSendAuditMessageAsync<TItem>(
```
robshakespeare: **todo:** Please can you remove this method from the interface, and make the method private in the class.

This ties in with a comment below.

Ultimately, the `AuditPepFindActivity` Azure Function can be removed now.

### Background:
`AuditPepFindActivity` is only used in the original fan-out process.  Because of the changes in this PR, in the fan-out process the auditing now happens internally (`SearchProviderSubOrchestrator` -> `FilterResultsByPolicyFunction` -> `IPolicyEnforcementAndAuditingService.FilterItemsAndAuditAsync`).  So by calling `AuditPepFindActivity` at the end of the fan-out process the results is now a duplicate audit entry (albeit a single larger one).  It is safe to remove `AuditPepFindActivity` because of this duplication, especially because it is only used in the fan-out process which is highly likely to be removed.

Apps/Find/src/SUI.Find.Application/Services/PepFilterAndAuditInput.cs
```diff
@@ -0,0 +1,25 @@
...
+/// <param name="Purpose">The reason why the Searcher/Requestor requested the data, and why they need to see the data.</param>
+/// <param name="InvocationId">ID used to correlate this action to the other steps in this search process.</param>
+public record PepFilterAndAuditInput<TItem>(
+    string SourceOrgId,
+    string DestOrgId,
+    string DestOrgType,
+    IReadOnlyList<TItem> Items,
+    DsaPolicyDefinition DsaPolicy,
+    string Purpose,
+    string InvocationId
```
robshakespeare: **nitpick:** Please can `InvocationId` be renamed to `CorrelationId`, to make less tied to Azure Functions and more to represent what it is ultimately used for.

Apps/Find/src/SUI.Find.FindApi/Functions/ActivityFunctions/AuditPepFindActivity.cs
```diff
@@ -1,25 +1,20 @@
...
 namespace SUI.Find.FindApi.Functions.ActivityFunctions;
 
 public record AuditPepFindInput(
     PolicyContext PolicyContext,
     SearchJobMetadata Metadata,
-    List<SearchResultWithDecision> SearchResultsWithDecisions
+    List<PepResultItem<CustodianSearchResultItem>> SearchResultsWithDecisions
 );
 
 public class AuditPepFindActivity(
```
robshakespeare: **todo:** It is safe to remove this whole Azure Function now, and remove the invocation from the `SearchOrchestrator`.  Please can you delete it in this PR.

### Background:
`AuditPepFindActivity` is only used in the original fan-out process.  Because of the changes in this PR, in the fan-out process the auditing now happens internally (`SearchProviderSubOrchestrator` -> `FilterResultsByPolicyFunction` -> `IPolicyEnforcementAndAuditingService.FilterItemsAndAuditAsync`).  So by calling `AuditPepFindActivity` at the end of the fan-out process the results is now a duplicate audit entry (albeit a single larger one).  It is safe to remove `AuditPepFindActivity` because of this duplication, especially because it is only used in the fan-out process which is highly likely to be removed.

Apps/Find/src/SUI.Find.Application/Interfaces/IPolicyEnforcementService.cs
robshakespeare, line 1: **todo:** I realise I suggested this class name, but now seeing how this whole set of changes has evolved, please can you keep this interface (and associated class) just being called `IPolicyEnforcementService`.

I think the method being called `FilterItemsAndAuditAsync` makes sense, and should be kept like that :+1:  But it makes sense to shorten and simplify the interface and class name, please.

Apps/Find/src/SUI.Find.Application/Services/PepFilterAndAuditInput.cs
```diff
@@ -0,0 +1,25 @@
...
+/// That service determines which of the specified items can be shared from the Source Organization (Custodian of data being shared) to the Destination Organization (Searcher/Requestor).
+/// </summary>
+/// <param name="SourceOrgId">Organization ID of the Custodian that owns the data being shared.</param>
+/// <param name="DestOrgId">Organization ID performing the search or making the request for the data, i.e. who the data is being shared to.</param>
+/// <param name="DestOrgType">Organization Type of the Searcher/Requestor.</param>
+/// <param name="Items">The items that represent the data being shared from the Source Organization to the Destination Organization.</param>
+/// <param name="DsaPolicy">The Data Sharing Agreement policy belonging to the Source Organization.</param>
+/// <param name="Purpose">The reason why the Searcher/Requestor requested the data, and why they need to see the data.</param>
+/// <param name="InvocationId">ID used to correlate this action to the other steps in this search process.</param>
+public record PepFilterAndAuditInput<TItem>(
```
robshakespeare: **todo:** Please can this just be called `PepFilterInput`.

Apps/Find/src/SUI.Find.Application/Services/PolicyEnforcementAndAuditingService.cs
```diff
@@ -1,21 +1,136 @@
...
+            cancellationToken
+        );
+
+        return resultsWithDecision;
+    }
+
+    public async Task CreateAndSendAuditMessageAsync<TItem>(
+        IReadOnlyList<PepResultItem<TItem>> resultsWithDecision,
+        string destinationOrgId,
+        string invocationId,
```
robshakespeare: **nitpick:** Please can `invocationId` be renamed to `correlationId`, to make less tied to Azure Functions and more to represent what it is ultimately used for.

Also, this method can be marked as `private` (please see related comments).

Apps/Find/src/SUI.Find.Infrastructure/Handlers/JobResultHandler.cs
```diff
@@ -218,19 +223,25 @@ CancellationToken cancellationToken
...
 
-        var resultsWithDecision = await pepService.FilterItemsAsync(
+        var filterInput = new PepFilterAndAuditInput<CustodianSearchResultItem>(
             context.Custodian.OrgId,
             context.SearchingOrganisation.OrgId,
             context.SearchingOrganisation.OrgType,
             records,
             context.Custodian.DsaPolicy,
             ApplicationConstants.PolicyEnforcementPurposes.Safeguarding,
+            invocationId
```
robshakespeare: **todo:** I think we should use the `WorkItemId` for correlation.  Ultimately the purpose of the correlation ID in the context of auditing is to tie the audit entries together to the search that they apply to.

This hadn't quite dawned on me until looking at this PR, but we need to use something different to the Invocation ID for the correlation of the audit entries.  The problem with using Invocation ID here is that it will differ from one result submission to the next within the same search, meaning that correlating the audit messages back to the same search will be difficult or maybe even not possible.  Hence I think `WorkItemId` is the right attribute to use for correlation in this context.

Apps/Find/src/SUI.Find.FindApi/Functions/QueueFunctions/JobResultHandlerFunction.cs
```diff
@@ -40,6 +40,6 @@ CancellationToken cancellationToken
             message.WorkItemId
         );
 
-        await handler.HandleAsync(message, cancellationToken);
+        await handler.HandleAsync(message, context.InvocationId, cancellationToken);
```
robshakespeare: **todo:** Related to my comment about audit correlation, I don't think InvocationId should be used, so there's no need to pass it in here.

Apps/Find/tests/SUI.Find.Application.UnitTests/Services/PolicyEnforcementServiceTests/EvaluateAsyncTests.cs
```diff
@@ -245,15 +247,20 @@ public async Task FilterResultsAsync_DoesEvaluateAsExpected()
...
 
-        // Act
-        var result = await _sut.FilterResultsAsync(
-            sourceOrgId: sourceOrgId,
-            destOrgId: destOrgId,
-            destOrgType: "LOCAL_AUTHORITY",
+        var filterInput = new PepFilterAndAuditInput<CustodianSearchResultItem>(
+            sourceOrgId,
+            destOrgId,
+            "LOCAL_AUTHORITY",
```
robshakespeare: **nitpick:** Please can the named arguments be kept for arguments that aren't obvious, especially because the constructor has many arguments and they're all similar/same types.

```suggestion
            destOrgType: "LOCAL_AUTHORITY",
```

Same for the `purpose` argument too, I think.

Apps/Find/tests/SUI.Find.Infrastructure.UnitTests/Handlers/JobResultHandlerTests.cs
```diff
@@ -392,13 +394,8 @@ await _jobService
...
             .Received(1)
-            .FilterItemsAsync(
-                message.CustodianId, // source (custodian)
-                searchingOrganisationId, // destination (searching org)
-                "TypeA",
-                Arg.Is<IReadOnlyList<CustodianSearchResultItem>>(x => x.Count == 2),
-                Arg.Any<DsaPolicyDefinition>(),
-                ApplicationConstants.PolicyEnforcementPurposes.Safeguarding,
+            .FilterItemsAndAuditAsync(
+                Arg.Any<PepFilterAndAuditInput<CustodianSearchResultItem>>(),
```
robshakespeare: **todo:** The refactor here has slightly reduced the attributes being verified.  Please can you change the assertion to be as specific as it was before.

So something like:

```
Arg.Is<PepFilterAndAuditInput<CustodianSearchResultItem>>(p =>
    p.SourceOrgId ==  message.CustodianId &&
    p.DestOrgId == searchingOrganisationId &&
    p.DestOrgType == "TypeA" &&
    p.Items.Count == 2 &&
    p.Purpose == ApplicationConstants.PolicyEnforcementPurposes.Safeguarding)
```

Thank you

robshakespeare (changes requested):
Thank you for taking the time to do this refactoring.  I can see was a complicated refactor, but it has been completed successfully and the end result is much cleaner.

I have added some comments, mostly additional tidy up.  Please can you do the changes?  I'm happy to go through the comments together on a call if that would help.  Many thanks.

Apps/Find/src/SUI.Find.FindApi/Functions/ActivityFunctions/AuditPepFindActivity.cs
robshakespeare: i.e. please remove the whole `AuditPepFindActivity` azure func.  Apologies, the comment hasn't come our exactly on the right line.  (But its the right file to be removed! 🙂)

Apps/Find/src/SUI.Find.Application/Interfaces/IPolicyEnforcementService.cs
robshakespeare: Thank you for updating that

Apps/Find/src/SUI.Find.Application/Models/JobContext.cs
robshakespeare: Thank you for updating this

Apps/Find/src/SUI.Find.Application/Interfaces/IPolicyEnforcementService.cs
robshakespeare: Thanks for making these updates

Apps/Find/src/SUI.Find.FindApi/Functions/ActivityFunctions/AuditPepFindActivity.cs
robshakespeare: Thank you for completing this

Apps/Find/src/SUI.Find.Application/Services/PolicyEnforcementAndAuditingService.cs
robshakespeare: Thank you

Apps/Find/src/SUI.Find.Infrastructure/Handlers/JobResultHandler.cs
robshakespeare: Thanks for making this update

Apps/Find/src/SUI.Find.FindApi/Functions/QueueFunctions/JobResultHandlerFunction.cs
robshakespeare: Thanks for updating this

Apps/Find/tests/SUI.Find.Application.UnitTests/Services/PolicyEnforcementServiceTests/EvaluateAsyncTests.cs
robshakespeare: Thanks for updating this

Apps/Find/tests/SUI.Find.Infrastructure.UnitTests/Handlers/JobResultHandlerTests.cs
robshakespeare: Many thanks for updating this

robshakespeare (approved):
Fantastic refactoring and updates.  Many thanks for doing these changes, and applying the feedback.

Neat result, especially for example how clean and easy to follow the `FilterItemsAndAuditAsync` method has ended up as :+1:

PR #573, opened 2026-04-20
docs: SUI-1606 - add auth environment strategy note

IgiCodes:
## Summary

This PR adds a new design note that defines how the inbound auth model should work across local development, CI, ephemeral environments, cheaper shared non-prod environments, and at least one shared high-fidelity environment.

The note builds on the existing [Authentication and API Edge Strategy](Docs/Design/Authentication/Index.md) and [Authentication Baseline and Security Model](Docs/Design/Authentication/BaselineSecurityModel.md) docs, and turns the environment strategy follow-on work into a concrete recommendation. It is intended to support implementation planning and auth hardening without treating every non-prod environment as if it must already be production-shaped.

## Changes

- Add a new design note at [Docs/Design/Authentication/EnvironmentStrategy.md](Docs/Design/Authentication/EnvironmentStrategy.md)
- Define the target auth shape for local, CI/ephemeral, cheaper shared non-prod, and shared high-fidelity environments, including the expected role of `APIM`, issuer approach by environment, and how seeded auth/test data should work
- Capture the minimum environment needed for useful auth learning, the key cost/fidelity trade-offs, and the recommendation to use real bearer-token flows rather than auth bypasses for realistic testing
- Link the new note from [Docs/Design/Index.md](Docs/Design/Index.md) and [Docs/index.md](Docs/index.md)

## Validation

- Docs-only change; no code/runtime validation required
- Reviewed against the current authentication design notes to keep the new environment strategy aligned with the existing auth baseline and immutable auth strategy direction

### DIFF
--- a/Docs/Design/Authentication/EnvironmentStrategy.md
+++ b/Docs/Design/Authentication/EnvironmentStrategy.md
@@ -1,0 +1,298 @@
+# Authentication Environment Strategy
+
+**Date:** `2026-04-14`  
+**Owner:** SUI Service Team  
+**Scope:** Target authentication environment strategy for local development, CI, ephemeral environments, cheaper shared non-prod environments, and at least one shared high-fidelity environment for the active inbound auth scope.
+
+This document builds on [Authentication and API Edge Strategy](./Index.md), [Authentication Baseline and Security Model](./BaselineSecurityModel.md), and [ADR-SUI-0011: Authentication and trust boundaries for SUI APIs](../../Architecture%20decisions/Systems%20landscape/0011-authentication-and-trust-boundaries-for-sui-apis.md).
+
+It defines how the auth baseline should be exercised across environments without relying on auth bypasses as a substitute for realistic testing.
+
+---
+
+## 1. Purpose
+
+The purpose of this note is to make the environment strategy concrete enough to support:
+
+- implementation planning for local, CI, and deployed non-prod auth
+- consistent auth behaviour across fast feedback and higher-fidelity environments
+- clear decisions on where `APIM` is required and where it is not
+- realistic testing using real tokens, scopes, and seeded identities instead of auth bypasses
+
+This note is an environment strategy, not a final production decision record.
+
+---
+
+## 2. Design Drivers
+
+The environment strategy is built around the following drivers:
+
+- the external auth contract should stay recognisably the same across environments
+- local and CI flows must stay fast and self-contained
+- at least one shared environment must exercise the likely Azure-hosted auth shape
+- cheaper shared environments should optimise for cost and regression coverage, not full edge fidelity
+- the services must remain issuer-agnostic and must not depend on provider-specific claim handling
+- tests that claim to cover auth should obtain and use real bearer tokens rather than bypass middleware or endpoint checks
+
+For the active scope, the stable cross-environment contract is:
+
+- `OAuth 2.0 client_credentials`
+- bearer `JWT` access tokens
+- scope-based API permissions
+- application-level authorisation using canonical organisation identity plus permissions
+
+---
+
+## 3. Current Starting Point
+
+The repo currently behaves as follows:
+
+- local development and ephemeral CI run the `Find` and `StubCustodians` services directly, without `APIM`
+- the `Find` API issues sandbox inbound tokens at `/api/v1/auth/token` using `Data/auth-clients-inbound.json`
+- the stub custodians issue sandbox outbound tokens at `/api/v1/auth/token` using `auth-clients-outbound.json`
+- `Find` validates bearer JWTs in application middleware today; there is no implemented gateway-trust mode yet
+- shared `d01` and `d02` environments currently use direct Function App and Web App URLs, not an `APIM` front door
+- shared auth and organisation fixture data now lives in the top-level `Data/` directory and is copied into app outputs
+
+This is already better than an auth bypass because tests obtain signed bearer tokens and hit protected endpoints. However, it is still a prototype environment shape rather than the full target operating model.
+
+Two current limitations matter for this strategy:
+
+- the shared hosted environments do not yet exercise `APIM` or an external issuer
+- the seeded inbound auth clients currently all carry the full scope set, which is useful for happy-path testing but weaker than the desired role-separated test model
+
+---
+
+## 4. Target Environment Shapes
+
+### 4.1 Environment matrix
+
+| Environment | Target edge shape | Target issuer shape | Validation and enforcement path | Seeded data approach | Primary purpose |
+| --- | --- | --- | --- | --- | --- |
+| Local development | Direct service access by default; optional local reverse proxy only for targeted gateway tests | Repo-managed sandbox issuer | `Find` validates JWTs in-app; application enforces scopes, ownership, and policy hooks | Repo-managed synthetic fixtures, including org directory and auth clients | Fast development and debugging |
+| CI ephemeral | Direct self-contained service startup; no `APIM` dependency | Same sandbox issuer contract as local | Same in-app JWT validation and endpoint protection as local | Same synthetic fixtures, resettable on every run | Repeatable automated regression and E2E testing |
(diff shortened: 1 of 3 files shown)


Docs/Design/Authentication/EnvironmentStrategy.md
```diff
@@ -0,0 +1,298 @@
...
+| --- | --- | --- | --- | --- | --- |
+| Local development | Direct service access by default; optional local reverse proxy only for targeted gateway tests | Repo-managed sandbox issuer | `Find` validates JWTs in-app; application enforces scopes, ownership, and policy hooks | Repo-managed synthetic fixtures, including org directory and auth clients | Fast development and debugging |
+| CI ephemeral | Direct self-contained service startup; no `APIM` dependency | Same sandbox issuer contract as local | Same in-app JWT validation and endpoint protection as local | Same synthetic fixtures, resettable on every run | Repeatable automated regression and E2E testing |
+| Cheaper shared non-prod | Direct service access by default; optional low-cost ingress later, but no `APIM` requirement | Environment-owned sandbox or test issuer preserving the same JWT contract | Application-level JWT validation remains authoritative | Same synthetic org and client catalogue, but shared-environment secrets should come from environment-managed secret storage | Hosted integration testing and lower-cost shared regression |
```
robshakespeare: **suggestion:** I'm wondering about this, and maybe it is fine, but is there a different or more specific term we can use instead of "Hosted integration testing".  Or are we referring to integration tests here because the purpose of them is to specifically test the auth (and hence shouldn't be classed as end-to-end tests per se)?

Docs/Design/Authentication/EnvironmentStrategy.md
```diff
@@ -0,0 +1,298 @@
...
+| Shared high-fidelity | `APIM` as mandatory ingress | Standards-compliant shared issuer, preferably `Entra` where governance allows | `APIM` performs coarse token checks; application still enforces permissions, ownership, and policy, and should continue JWT validation until a trusted gateway-forwarded auth path exists | Same synthetic participants may still be used for learning, but onboarding, credentials, and signing material should be environment-managed | Operational learning on issuer, gateway, onboarding, and auth support model |
```
robshakespeare: **suggestion:** instead of

> and should continue JWT validation until a trusted gateway-forwarded auth path exists

can we not just say

> and should continue JWT validation

I like the idea that the app remains authoritative always, and that would be useful throughout the project even as it becomes stabilised.

Docs/Design/Authentication/EnvironmentStrategy.md
```diff
@@ -0,0 +1,298 @@
...
+### 4.4 Cheaper shared non-prod environments
+
+The cheaper shared non-prod target shape is:
+
+- keep direct hosted service access as the default
+- do not require `APIM` in every shared dev/test environment
+- keep JWT validation and scope enforcement inside the application
+- preserve the same token and scope contract used locally and in CI
+
+Current `d01` and `d02` fit this bucket more naturally than the high-fidelity bucket. They are useful for hosted regression and integration testing, but they are not yet sufficient for learning the final issuer and gateway operating model.
```
robshakespeare: **nitpick:** Should this say "Currently"?

```suggestion
Currently `d01` and `d02` fit this bucket more naturally than the high-fidelity bucket. They are useful for hosted regression and integration testing, but they are not yet sufficient for learning the final issuer and gateway operating model.
```

Docs/Design/Authentication/EnvironmentStrategy.md
```diff
@@ -0,0 +1,298 @@
...
+## 10. Current Recommendation
+
+The current recommended target is:
+
+- local development uses direct service access, sandbox token issuance, and application-level JWT validation
+- CI and ephemeral environments use the same auth contract and avoid test-only bypass paths
+- current shared `d01` and `d02` environments continue to act as cheaper shared non-prod environments, not as the primary high-fidelity auth environment
+- at least one shared high-fidelity environment should be introduced or designated for `APIM` plus shared-issuer learning
+- seeded auth data should be tightened to model searcher-only, custodian-only, dual-role, and negative-test clients more realistically
+- shared-environment secret material should move out of repo-managed shared fixtures and into environment-managed secret storage or issuer-managed configuration
```
robshakespeare: **question:** To me this feels like it should only concern the 'Cheaper shared non-prod' environments, rather than also effecting Local and CI.  Do you think it is ok for Local and CI and stay using repo secrets?  My angle on this is to reduce the manual configuration required for local developer setup.  Ideally repos should be able to be pulled down and run without any manual setup, aside from installing prerequisite tooling.  (And it is fine to have public secrets for purely local / CI development, because the secrets don't actually protect anything, they are just used so the auth remains real for local / CI, rather than bypassed).

Docs/Design/Authentication/EnvironmentStrategy.md
```diff
@@ -0,0 +1,298 @@
...
+---
+
+## 10. Current Recommendation
+
+The current recommended target is:
+
+- local development uses direct service access, sandbox token issuance, and application-level JWT validation
+- CI and ephemeral environments use the same auth contract and avoid test-only bypass paths
+- current shared `d01` and `d02` environments continue to act as cheaper shared non-prod environments, not as the primary high-fidelity auth environment
+- at least one shared high-fidelity environment should be introduced or designated for `APIM` plus shared-issuer learning
```
robshakespeare: **question:** Please can I just double check what is meant by "shared" specifically in the context of shared-issuer?

Docs/Design/Authentication/EnvironmentStrategy.md
IgiCodes: Good point. I've changed that wording to "Hosted regression testing and lower-cost shared validation" to avoid overloading "integration testing" there.

Docs/Design/Authentication/EnvironmentStrategy.md
IgiCodes: Agreed. I've simplified that line so it now just says the application should continue JWT validation.

I think that is the clearer statement of the intended baseline anyway, because it keeps the app authoritative even when APIM is present.

Docs/Design/Authentication/EnvironmentStrategy.md
IgiCodes: Yes, agreed. I've updated that sentence to start with "Currently" so it reads as a present-state observation rather than a permanent classification.

Docs/Design/Authentication/EnvironmentStrategy.md
IgiCodes: I've clarified that a bit better in the doc.

What I meant by "shared issuer" was an issuer that is centrally managed for a deployed shared environment, rather than the repo-managed sandbox issuer to keep local and CI self-contained.

Docs/Design/Authentication/EnvironmentStrategy.md
IgiCodes: Yes, I agree with that distinction.

I've tightened the wording so the move away from repo-managed secret material applies to shared deployed environments, while local/CI/ephemeral can remain repo self-contained with repo-managed synthetic credentials and signing material where needed to keep setup friction low.

So the intent is still:
- local and CI stay easy to pull down and run
- shared environments move toward environment-managed secret material

robshakespeare: Thanks, yes agreed, that is clearer now, good approach.

Docs/Design/Authentication/EnvironmentStrategy.md
robshakespeare: Thank you for making that clearer :+1:

Docs/Design/Authentication/EnvironmentStrategy.md
robshakespeare: Ace, thank you

Docs/Design/Authentication/EnvironmentStrategy.md
robshakespeare: Great thank you for understanding and the update here

robshakespeare (approved):
Excellent document, and thank you for working through the feedback.

I think this document sets out a clear and sensible way forward.

PR #577, opened 2026-04-22
feat: SUI-1671 - improve e2e test reliability immediately after deployment of the Stub Custodians

robshakespeare:
## Summary
This is the same change as commit 0a5599a1470f39169bd904a03808f128c3c55a04, but just for the Stubs rather than Find.  It updates the e2e tests to wait for the latest version of the Stubs to be deployed and started before the tests commence.

## Changes
The changes in this PR ensure that the e2e tests wait for the new version of the Stubs to be deployed and up and running.  Rather than wait for an arbitrary period of time (which could wait for too long unnecessarily), it waits until the API responds with a build timestamp which is on or after a known point in time that confirms the latest version is in use.

### Context and background info:
The e2e tests falsely fail when they are automatically ran after deployment of the Stubs.  They fail because the e2e tests run immediately after deployment, at which point the previously deployed version of the Stubs reports that is is up and healthy because the new deployment hasn’t started yet.  Once the new version of the Stubs is deployed and restarts, the API then temporarily times-out, and after a non-determinate period of time the new version of the app is up and healthy.   By waiting for this new version to be up before running the e2e tests, we ensure we are running against the latest version and that we won't get false failures from temporary timeouts caused by the deployment.

## Validation
- Unit, integration, E2E tests complete successfully locally
- E2E tests workflow passes against ephemeral environment, as part of PR validation

### DIFF
--- a/.github/workflows/find-e2e-tests.yml
+++ b/.github/workflows/find-e2e-tests.yml
@@ -181,8 +181,8 @@ jobs:
             -e E2E__BaseUrl=http://localhost:7182/api/ \
             -e E2E__StubCustodiansBaseUrl=http://localhost:5193/api/ \
             -e E2E__SkipResetAzureTables=False \
-            -e E2E__UseExtendedFindApiHealthCheckTimeout=False \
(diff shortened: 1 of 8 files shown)


udara-jay (approved):
Looks great! Thank you for continuously looking into this 👍

PR #580, opened 2026-04-24
SUI-1612 - Add PEP filtering before creating jobs

tomhawkin:
## Changes
added PEP filtering to the SearchQueueJobTrigger. if PEP filtering returns a denial then we don't create a job for that custodian

- added PEP to SearchQueueJobTrigger
- added method to Custodian Service to find organisation in a specified list, so that the case sensitively checks are consistent

## Validation

- created unit tests
- manual validation


Apps/Find/src/SUI.Find.FindApi/Functions/QueueFunctions/QueueSearchJobTrigger.cs
```diff
@@ -49,9 +50,36 @@ CancellationToken token
         );
 
         var custodians = await custodianService.GetCustodiansAsync();
+        var requestingOrg = custodians.First(s =>
```
robshakespeare: **todo:** The ICustodianService already has this functionality, please can you change this to use `ICustodianService.GetCustodianAsync(searchRequestMessage.SearchingOrganisationId)`

Apps/Find/src/SUI.Find.FindApi/Functions/QueueFunctions/QueueSearchJobTrigger.cs
```diff
@@ -49,9 +50,36 @@ CancellationToken token
...
+            s.OrgId == searchRequestMessage.SearchingOrganisationId
+        );
+        var createdJobs = 0;
 
         foreach (var custodian in custodians)
         {
+            var pepDecision = await policyEnforcementService.EvaluateAsync(
```
robshakespeare: **todo:** Now looking at how this has come together, I think we should *not* change the PolicyEnforcementService.  Each "custodian" here is actually a provider definition, which includes the record type, so I think we should just pass the record type along in the call here.

Also, the plan was to use the new `FilterItemsAndAuditAsync` method, because we must audit the PEP activity.  It does this say this in the ticket, but I realise it could have been clearer, my apologies.

Please can you do the following changes:
* Revert the changes to `PolicyEnforcementService` and `PolicyDecisionRequest` in this PR, so that record type remains mandatory for PEP.
* Change to use the `FilterItemsAndAuditAsync` method here instead.  I think we should group the list of `custodians` (i.e. provider definitions) by `OrgId` so that we only call the PEP service once per "source" organisation.
* Include `custodian.RecordType` in the request.

I'm happy to have a call to chat this through if that would help?  Many thanks

robshakespeare (changes requested):

Apps/Find/src/SUI.Find.FindApi/Functions/QueueFunctions/QueueSearchJobTrigger.cs
tomhawkin: i wrote it like this because it was more efficient to get the correct one of out the current list than go back and load the whole list up again to get one out. I can change it if you want but it will be slightly less efficient

robshakespeare: Ah ok, that's fair.  Leave it for now, I'll have a think about it.  Thanks

Apps/Find/src/SUI.Find.Application/Services/PolicyEnforcementService.cs
```diff
@@ -210,7 +210,7 @@ private static bool RuleMatches(
         PolicyDecisionRequest request,
         string destOrgType,
         string modeString,
-        string recordType,
+        string? recordType,
```
robshakespeare: **todo:** I think this change can be reverted?

Apps/Find/src/SUI.Find.FindApi/Functions/QueueFunctions/QueueSearchJobTrigger.cs
robshakespeare: Many thanks for following on with these updates

Apps/Find/src/SUI.Find.FindApi/Functions/QueueFunctions/QueueSearchJobTrigger.cs
```diff
@@ -49,8 +52,31 @@ CancellationToken token
...
+            s.OrgId == searchRequestMessage.SearchingOrganisationId
+        );
+        var createdJobs = 0;
+        var pepFilteredCustodians = new List<PepResultItem<ProviderDefinition>>();
 
         foreach (var custodian in custodians)
+        {
+            pepFilteredCustodians.AddRange(
+                await policyEnforcementService.FilterItemsAndAuditAsync(
+                    new PepFilterInput<ProviderDefinition>(
```
robshakespeare, line 66: **nitpick:** Please can you update this to use named arguments.  The constructor is a big one, which isn't ideal, so specifying the named arguments helps to reason with and understand the code.

Something like:
```
new PepFilterInput<ProviderDefinition>(
	SourceOrgId: custodian.OrgId,
	DestOrgId: requestingOrg.OrgId,
	DestOrgType: requestingOrg.OrgType,
	Items: [custodian],
	custodian.DsaPolicy,
	Purpose: "SAFEGUARDING",
	CorrelationId: searchRequestMessage.WorkItemId.ToString()
)
```

Apps/Find/tests/SUI.Find.Application.UnitTests/Services/PolicyEnforcementServiceTests/EvaluateAsyncTests.cs
```diff
@@ -377,4 +377,61 @@ public async Task FilterItemsAsync_DoesEvaluateAsExpected()
...
+                    DestOrgTypes = ["POLICE"], // Only POLICE, so LOCAL_AUTHORITY will be denied
+                    Purposes = ["SAFEGUARDING"],
+                },
+            ],
+        };
+
+        var request = new PolicyDecisionRequest(
+            SourceOrgId: "MULTIPLE-01",
+            DestinationOrgId: "LOCAL-AUTHORITY-01",
+            RecordType: null,
```
robshakespeare: **question:** Are these tests valid now?

robshakespeare (commented):
Many thanks for following on with these updates.  Just a few more suggestions and a question, thank you.

Apps/Find/src/SUI.Find.FindApi/Functions/QueueFunctions/QueueSearchJobTrigger.cs
robshakespeare: Thanks for updating that 👍

robshakespeare (approved):
Looks good, thank you

PR #588, opened 2026-04-29
fix: SUI-1670 - resolved intermittent failure to load PDS credentials in Match function when running locally

robshakespeare:
## Summary

This PR fixes intermittent errors experienced when running the Match function locally, because the NHS config values (PDS credentials) weren't loading correctly.  Fixes SUI-1670.

## Changes

- fix: removed conflicting lines from the sample local settings file.  The expected setup is to have them defined in the `.env` file instead (as per the readme).  Having them here too causes the values to be intermittently loaded from here, rather than correctly from the `.env` file, causing intermittent failures of Match when running locally.
- docs: improved the `pds_fhir_local_setup.md` readme, to help with resolving intermittent errors when the config values aren't loading correctly.

Also includes:
- feat: added .NET User Secrets support to the FindApi.  Not necessary, but its useful and is a straight forward single line change.
- refactor: standardised the existing .NET User Secrets setup for the E2E tests (Changes folder from: `~/.microsoft/usersecrets/SUI.E2E.Tests/secrets.json` to: `~/.microsoft/usersecrets/SUI.Find.E2ETests/secrets.json`.

## Validation

- Manual verification by running the Match function locally, and confirming it consistently works


udara-jay (approved):
Looks great! 👍

PR #589, opened 2026-04-29
SUI-1697 - added password to test harness

tomhawkin:
## Summary

added the ability to set the password for the test harness in config. Incorrect password now blocks login.
## Changes

- added password to config file
- added logic to check password

## Validation

- run the test harness, try to log in with incorrect password


robshakespeare (approved):
Nice :+1:

PR #590, opened 2026-04-29
test(find): SUI-1669 - Test e2e parallelisation with xunit v3

robshakespeare:
## Summary

Updated the Find E2E Tests to enable parallelisation.  In local test runs, this reduced execution times down from 15 minutes to 9 minutes.  Relates to SUI-1669.

## Changes

- Refactoring and updates to enable thread safety, so that the tests can reliably run in parallel.
- Upgraded to xUnit v3 so that we can use `AssemblyFixture` for a global setup class, ultimately so that test classes can be in different collections, to enable them to run in parallel.
- Separated out the happy path & no-records path tests, to increases parallelisation.
- Added rudimentary script for reliably running the e2e tests locally (this mostly removes the false failures caused by Rider, and provides a repeatable approach to get fair test execution times)

## Validation

- Local e2e test runs


robshakespeare (commented):
I'm happy for this PR to be merged now.

Apps/Find/tests/SUI.Find.E2ETests/xunit.runner.json
```diff
@@ -1,5 +1,8 @@
 ﻿{
   "$schema": "https://xunit.net/schema/current/xunit.runner.schema.json",
   "showLiveOutput": true,
-  "diagnosticMessages": true
+  "diagnosticMessages": true,
+  "parallelizeAssembly": true,
+  "parallelizeTestCollections": true,
```
udara-jay, line 6: All looking fantastic, thank you for all the improvements and updates! 🔥 

Just wondering if we need this in the config anymore since we don't have any collections now. i.e `parallelizeTestCollections `

robshakespeare: Good question, I just checked, each class becomes a collection by default, just explicitly declaring collections using attributes allows multiple classes to be grouped in to a single collection.  But otherwise, each class just becomes a collection, and the collection is the unit that gets parallelised.

So I think we should keep this, just to be clear if anything (because I think its the default).

udara-jay: All good, no change needed based on our chat 👍

robshakespeare: Thanks 🙂

udara-jay (approved):
Glad everything worked in the end and to see the reduction in execution time! Thank you for your input on this 👍

PR #592, opened 2026-04-30
feat: SUI-1665: update UI Test Harness to use the auth-clients-inbound.json approach

robshakespeare:
## Summary

Updates the UI Test Harness to dynamically include the registered Custodians, and their authentication details; so that there are less hard coded places to update, and to ease migration to a more production ready Authentication approach.  Relates to SUI-1665.

## Changes

- Updated to use the `Data/auth-clients-inbound.json` file and approach, like is done already in the `StubCustodians`.
- Updated the UI Test Harness’ list of “Custodians to simulate” so that it is populated from the data file.
- Updated the UI Test Harness’ token generation ultimately uses the credentials from the data file.
- Updated relevant GitHub workflows to be triggered on the `Data` directory too.

Also includes related fixes/improvements:
* Updated UI Test Harness to include loading spinners, to indicate work is being done (the progress bar can stay at 0% for a while, especially when using the Polling architecture, so the spinner provides more confidence that something is being done).
* Fixed the UI Test Harness' Find exit condition to stop when the percentage reaches 100% or the job becomes completed, rather than being capped at 10 steps.  Fixes issue whereby using the Polling Architecture would not correctly find all the expected records, because it usually takes longer than 10 steps.
* Fixed the UI Test Harness's Find Records process so that it breaks out if the NHS number changes.  Fixes issues whereby the background async process keeps running even if the user switches tab and chooses a different person, which would previously incorrectly cause records from the old process to be added to the new process.

## Validation

- Manual UI validation and testing.

### DIFF
--- a/.github/workflows/custodians-build-and-deploy.yml
+++ b/.github/workflows/custodians-build-and-deploy.yml
@@ -45,11 +45,13 @@ on:
       - main
     paths:
       - "Apps/StubCustodians/**"
+      - "Data/**"
   pull_request:
     branches:
       - main
     paths:
       - "Apps/StubCustodians/**"
+      - "Data/**"
 
 jobs:
   build-and-test-StubCustodians:
--- a/.github/workflows/find-build-and-deploy.yml
+++ b/.github/workflows/find-build-and-deploy.yml
@@ -45,11 +45,13 @@ on:
       - main
     paths:
       - "Apps/Find/**"
+      - "Data/**"
   pull_request:
     branches:
       - main
     paths:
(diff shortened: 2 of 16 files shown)


udara-jay (approved):
Looks great! 
NO-DSA-01 is a much better name option than FAUX-01 👍

robshakespeare:
> Looks great! NO-DSA-01 is a much better name option than FAUX-01 👍

Thank you.

Yes, does what it say on the tin then!  Faux was a poor name, my bad! 🙂

PR #595, opened 2026-04-30
added error message to login screen when incorrect password is entered

tomhawkin:
## Summary
added a redirect back to the login page with query parameters if the password is incorrect

## Changes

- password logic to redirect back to the login page
- login page changes to show error message

## Validation

- run ui test harness and log in with incorrect password


robshakespeare (approved):
Ace :+1:

PR #596, opened 2026-05-01
feat(find): SUI-1695 - Add backoff signalling to no work available responses

udara-jay:
## Summary

The ‘No work available' responses in the Polling implementation should include Backoff Signalling, so that load to the service can be dynamically managed in the future.

## Changes

In the 204 NoContent responses from both the ClaimJobFunction and the WorkAvailableFunction include a Retry-After response header, with a hardcoded value of 10 (seconds).

The generated OpenAPI schema is updated to indicate that the Retry-After response header will be included for 204 NoContent responses.

## Validation

Unit Tests with updates

### DIFF
--- a/Apps/Find/src/SUI.Find.FindApi/Functions/HttpFunctions/ClaimJobFunction.cs
+++ b/Apps/Find/src/SUI.Find.FindApi/Functions/HttpFunctions/ClaimJobFunction.cs
@@ -84,7 +84,7 @@ public class ClaimJobFunction(ILogger<ClaimJobFunction> logger, IJobClaimService
         }
 
         var response = HttpResponseUtility.NoContentResponse(req);
-        response.Headers.Add("Retry-After", "10");
+        response.Headers.Add("Retry-After", ApplicationConstants.Http.DefaultRetryAfterSeconds);
         return response;
     }
 }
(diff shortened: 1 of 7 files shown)


Apps/Find/src/SUI.Find.FindApi/Functions/HttpFunctions/ClaimJobFunction.cs
```diff
@@ -81,6 +83,8 @@ CancellationToken cancellationToken
             return await HttpResponseUtility.CreatedResponse(req, claimedJob, cancellationToken);
         }
 
-        return HttpResponseUtility.NoContentResponse(req);
+        var response = HttpResponseUtility.NoContentResponse(req);
+        response.Headers.Add("Retry-After", "10");
```
robshakespeare: **nitpick:** Please could you pull the `"10"` value out in to a constant, and then update the 2 places to use that constant.  Thank you.

robshakespeare (changes requested):
Nicely done, just one nitpick in regards to have a constant value for the 2 places.  This will help us find the places to update when we come to upgrade this in the future.  Thanks

robshakespeare (approved):
Brill, thanks

PR #599, opened 2026-05-05
feat: SUI-1579 - Rework approach to storing NHS FHIR secrets

jordanstarksbrowning-hippo:
## Summary

Currently the following secrets are stored in key vault and are accessed directly via an async security service:

- nhs-digital-private-key
- nhs-digital-kid
- nhs-digital-client-id 

In order to more easily facilitate the deployment of these secrets to multiple environments, this PR reworks the flow of information to be the following:

Github Secrets (updated manually) -> Azure Key Vault (via terraform env variables) -> .NET config file (via Key Vault References) -> Program (via AuthTokenServiceConfig)

## Changes

- Inject the github secrets into terraform via the plan-and-apply workflow
- Add the secrets to Key Vault via the find terraform files, and link them to the relevant app settings
- Refactor FhirAuthTokenService to read the secrets directly from the config settings instead of the AzureKeyVaultSecretService
- Remove all code relating to the AzureKeyVaultSecretService
- Remove redundant testing code and package references

## Validation

- Unit/integration/E2E tests run
- Match Service smoke test run
- Terraform plan/apply on d01 and d02


.github/workflows/terraform-plan-and-apply.yml
```diff
@@ -25,6 +25,13 @@ on:
         required: true
       AZURE_SUBSCRIPTION_ID:
         required: true
+      NHS_DIGITAL_CLIENT_ID:
+        required: true
```
robshakespeare, line 30: **suggestion:** I think we should make the 3 new secrets `required: false`.  The reason being, they're only needed by the Match endpoint.  My suggestion is to change this code to be:

```
      # The NHS secrets are only required by the Match endpoint
      NHS_DIGITAL_CLIENT_ID:
        required: false
      NHS_DIGITAL_KID:
        required: false
      NHS_DIGITAL_PRIVATE_KEY:
        required: false
```

What do you think?  I think it is a good idea to show the intent more clearly here, so that we remember when we come to look at this in the future.  It is most likely that Match is removed from here, with it probably moving to the NHS estate.

The alternative would be to make this all dynamic, and pass in a mapping or matrix, but I think that is super complex and would actually make it harder to understand.  So my suggestion here is just a pragmatic approach to indicate the specific uses of these values in a generic template file.

robshakespeare (commented):
This is looking fantastic, thank you for doing this.

Just one thing I am wondering about.

.github/workflows/terraform-plan-and-apply.yml
jordanstarksbrowning-hippo: I understand the thinking here, my concern would be that if these secrets are not present in github, would an empty value be transmitted to the key vault?

I also can't think of a situation where these values wouldn't be present in github, except for the point where the match service is moved, and in that case these secrets would be removed anyway.

Definitely happy to add the comment to specify why these secrets are present, but I'm not sure what value we would bring by allowing these to be removed from github when we would still need them in azure.

robshakespeare: Coolio, yes good point.  I'm happy to just add the comment :+1:  Please can you do that?

.github/workflows/terraform-plan-and-apply.yml
```diff
@@ -25,6 +25,14 @@ on:
         required: true
       AZURE_SUBSCRIPTION_ID:
         required: true
+      # The below secrets are required for the Match Function.
```
robshakespeare: **nitpick:** Sorry to be super picky!

```suggestion
      # The NHS secrets below are only required for the Match Function.
```

robshakespeare (approved):
Great, thanks

PR #607, opened 2026-05-07
feat: SUI-1708 - restart Test Harness on Match API Key rotation

robshakespeare:
## Summary

When the Match API Key is rotated (which is scheduled to automatically happen every week) we need to restart the UI Test Harness webapp, so that it picks up and uses the latest API Key for its calls to the Match function.

## Changes

- Updated the placeholder `restart-webapp.yml‎` workflow to be fully working.  This workflow restarts the specified webapp, and can be invoked manually or as a template from another workflow.
- Updated the existing `rotate-keyvault-secret.yml` workflow (and related files) to restart `uiharness01` when the `match-api-key` is rotated.

Also includes:
- Fixed issue in `rotate-keyvault-secret.yml‎` introduced because of the need to now have the NHS secrets available as TF vars.
- Minor update of existing comment in `terraform-plan-and-apply.yml`

## Validation

- Manual workflow runs


udara-jay (approved):
Looks good to me 👍

PR #612, opened 2026-05-08
Ops: SUI-1678 fixes

udara-jay:
fixes for getting ui harness deployed


robshakespeare (dismissed):
Nice 👍

robshakespeare (approved):
Neat follow-on solutions, especially given that this is only a test harness.

PR #618, opened 2026-05-12
chore: SUI-1754 - Sprint 9 housekeeping

robshakespeare:
## Summary

This PR tidies up some minor hangovers from recently completed work.

## Changes

- Removed the old `KeyVaultUri` and `UseStubAuthTokenService` config values, they are no longer used. (These values are no longer used now that the PDS FHIR secrets are always loaded from .NET config, and they are securely provided via Azure App Settings using Key Vault References in the infrastructure.)
- The UI Test Harness should include the Data folder in its workflow triggers, because it uses the shared Data to populate its list of Custodians, and get the Client Secrets to obtain auth tokens.

## Validation

- Manual checks & verification
- Planned the Terraform changes, to see that the only actual change is to remove the config values that are no longer used.
- Applied the Terraform changes to `d01` and `d02`

### DIFF
--- a/.github/workflows/ui-harness-build-and-deploy.yml
+++ b/.github/workflows/ui-harness-build-and-deploy.yml
@@ -45,11 +45,13 @@ on:
       - main
     paths:
       - "Apps/UIHarness/**"
+      - "Data/**"
   pull_request:
     branches:
       - main
     paths:
       - "Apps/UIHarness/**"
+      - "Data/**"
 
 jobs:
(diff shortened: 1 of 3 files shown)


udara-jay (approved):
Looks good and thank you for adding missing bits to the ui-harness as well 👍

robshakespeare:
> Looks good and thank you for adding missing bits to the ui-harness as well 👍

Ace, thanks 🙂

PR #619, opened 2026-05-13
feat: SUI-1677 - Update the generated 'Get an Identifier' OpenAPI spec

udara-jay:
## Summary

Update the generated 'Get an Identifier' OpenAPI spec to match the polished version, so that learn and develop patterns for providing polished, publish-ready OpenAPI specs.

## Changes

Endpoint Updates: Summary updated, match-record.read scope enforced, and the JSON Request Body Example is wired up.

Response Descriptions: 200, 400, 401, 404, and the brand new 500 error are fully documented.

Schema Updates: MatchRequest, PersonSpecification, MatchRequestMetadata, PersonMatch, and Problem schemas all have the exact descriptions, required arrays, and examples requested.

The DateOnly Bug: Fixed via the FindDocumentFilter.

## Validation

Made sure that unit tests still pass.

Run Find API and check for updates via the swagger UI.


Apps/Find/src/SUI.Find.FindApi/Models/Problem.cs
```diff
@@ -1,3 +1,29 @@
...
+    )]
+        string Title,
+    [property: OpenApiProperty(
+        Description = "The HTTP status code generated by the origin server for this occurrence of the problem; see RFC 9110, Section 15.",
+        Nullable = false
+    )]
+        int Status,
+    [property: OpenApiProperty(
+        Description = "A human-readable explanation specific to this occurrence of the problem.",
+        Nullable = false
```
robshakespeare: **suggestion:** I wonder whether `Nullable = false` is correct for the `Detail` prop?  I realise this probably contradicts what I specified in the ticket, but now looking at the code I think it should be `Nullable = true` (or omitted) for the `Detail` prop.

What do you think?

robshakespeare (commented):
Fantastic set of changes.  Great learnings and discoveries regarding how best to control the generated OpenAPI spec.  Great fix for `dateOnly` too.  Thanks :+1:

Apps/Find/src/SUI.Find.FindApi/Models/Problem.cs
udara-jay: I think yes, it should be changed or removed. Good catch 👍

robshakespeare (approved):
Thanks for the thorough approach to this :+1:

PR #622, opened 2026-05-14
feat: SUI-1751 - UI Test Harness: check for find results more frequently

robshakespeare:
## Summary

The changes in this PR update the UI Test Harness to check for find-a-record results more frequently, so that the UI does not look like it is hanging and so that results don’t appear in the UI in large batches.  Ultimately this will improve the demonstration of the system.

## Changes

- Changes to check for find-a-record results more frequently.
- Improved the fanout process so its straight forward to change its maximum time to wait.
- Updates to introduce and Architecture Constants rather than magic strings.

## Validation

- Manually tested via the UI locally


udara-jay (approved):
Looks great! I will incorporate these into mine 👍

robshakespeare:
> Looks great! I will incorporate these into mine 👍

Brilliant, thank you 🙂

PR #623, opened 2026-05-14
tests: SUI-1615 - E2E Test: Cross-organisation isolation for searches

robshakespeare:
## Summary

Adds a set of E2E tests to verify that an Organisation cannot access other Organisation’s searches.

## Changes

- Added new E2E test, using existing pattern, that starts a search as one client, then proves another client cannot read status, results, fetch the returned records, or cancel the job.
    - Also includes E2E test change that only effects the version 1 "start a new search tests": now that there are two sets of version 1 "start a new search tests", they need to be put in to the same collection, because they cannot reliably run in parallel because of the shared job issue in version 1.

- Updated all HTTP endpoints in FindAPI to standardise on using `Forbidden`, rather than `Unauthorized`, when the request doesn't have access to the resource.

- Fixes to existing issues that were uncovered by these tests:
    - Fixed issue where by Fetch Record would incorrectly return a 500 server error rather than return 404 Not Found, when the record does not exist.
    - Fixed issue whereby v1 Jobs could never be cancelled, instead it always incorrectly threw the error: "The ReadInputAs method can only be used on OrchestrationMetadata objects that are fetched with the option to include input data."

## Validation

- E2E tests runs locally and ephemeral runs in GitHub

### DIFF
--- a/Apps/Find/src/SUI.Find.Application/Models/CustomOneOf.cs
+++ b/Apps/Find/src/SUI.Find.Application/Models/CustomOneOf.cs
@@ -1,3 +1,5 @@
 namespace SUI.Find.Application.Models;
 
 public struct Unauthorized;
+
+public struct Forbidden;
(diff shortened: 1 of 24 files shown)


udara-jay (approved):
Looks very clean and thank you for ironing out the underlying issues 👍

PR #624, opened 2026-05-15
feat: SUI - 1758 : UI Test Harness updates based on UR team feedback

udara-jay:
## Summary

Updates were requested to enable the Test Harness to be used in the research sessions.

As part of this, the researcher team want to remove the Architecture dropdown because it is too technical, and not something they need to test in the research.  However, its useful for testing purposes, while fanout still exists in the code.  So, we should make the visibility of the Architecture dropdown configurable.

## Changes

Text Changes: The disclaimer is on the login screen, all "Enrol" text has been updated to "Get NHS Numbers", and the Find Records table is cleaned up.

Architecture Config: The ArchitectureOptionHidden flag is wired up in the Blazor code to safely hide the dropdown and default to "Polling" when set to true.

Terraform: The d02 environment variables file is updated to flip that flag specifically for their environment.

## Validation

Locally tested with the UI Test Harness

### DIFF
--- a/Apps/UIHarness/src/SUI.UIHarness.Web/Components/Controls/FindTab.razor
+++ b/Apps/UIHarness/src/SUI.UIHarness.Web/Components/Controls/FindTab.razor
@@ -37,7 +37,8 @@
         <tr>
             <th>Custodian</th>
             <th>Record type</th>
-            <th></th> </tr>
+            <th></th> 
+        </tr>
         </thead>
         <tbody>
         @if (_foundRecordsFor?.Org == SelectedOrg && _foundRecordsFor?.NhsNumber == CurrentPerson?.NhsNumber)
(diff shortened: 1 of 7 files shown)


Apps/UIHarness/src/SUI.UIHarness.Web/Components/Controls/FindTab.razor
```diff
@@ -34,29 +34,25 @@
...
-            <tr>
-                <th>Custodian</th>
-                <th>Record type</th>
-                <th>Record ID</th>
-                <th>Action</th>
-            </tr>
+        <tr>
+            <th>Custodian</th>
+            <th>Record type</th>
+            <th></th> </tr>
```
robshakespeare: **nitpick (non-blocking):** Just one thing, and this is a super nitpick, but I'm wondering whether having the `</tr>` closing tag on the end of the `th` line is intentional?

robshakespeare (dismissed):
Nice set of changes.  Great idea putting the hidden input in there for when the dropdown isn't shown :+1:

robshakespeare (approved):
Super, thanks

PR #627, opened 2026-05-19
ops: sui-1746 - Create separate "Auxiliary" service plan for stubs and non-prod apps

udara-jay:
## Summary

Create a separate "Auxiliary" service plan for stubs and non-prod apps; so that the main app service plan is representative of how our system will run for real, and also so that the stubs, mocks, and demo apps do not interfere with the performance on the main apps (i.e. just Find API at the moment).  Ultimately this will ensure we get accurate readings on our Find API’s performance.

## Changes

New auxiliary ASP suffixed with `auxsvcs01` created in core.

- Optional via `use_auxiliary_asp` boolean
- `false` by default
- `d01` and `d02` set as true
- Created in the same region (UK West)
- Tier settings (B1, Linux, 1 worker)

## Validation

Plan terraform to d01 from branch

### DIFF
--- a/terraform/core/main.tf
+++ b/terraform/core/main.tf
@@ -27,6 +27,12 @@ locals {
     var.environment_id,
     var.region_short
   )
+  aux_asp_name = format(
+    "%s%sasp-%s-auxsvcs01",
+    var.subscription_prefix,
+    var.environment_id,
+    var.region_short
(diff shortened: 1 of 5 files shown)


robshakespeare (approved):
Looks spot on, thanks

PR #632, opened 2026-05-20
feat(find): sui-1745 - Cache loaded org directory in FindAPI's Custodian Service

udara-jay:
## Summary

Cache the loaded org directory in MockCustodianService, so that the frequently called polling endpoints do not unnecessarily slow down the system.

## Changes

Memory caching should be a straight forward hold in an lazy in-memory list, following the pattern in the Stub’s FindApiAuthClientProvider:
- The logic to load the ‘provider definitions’ is moved in to a Lazy<IReadOnlyList<ProviderDefinition>> that is initialised in the class constructor
- The classes methods are refactored to use that in-memory list
- Update tests

## Validation

Validated by local and pipeline tests (unit, integration and E2E)


robshakespeare (approved):
Great, thank you for doing these changes.

PR #634, opened 2026-05-21
feat(auth): SUI-1749 - Create Auth Emulator app

jordanstarksbrowning-hippo:
## Summary

This PR creates a new self-contained ASP.NET web app which contains the functional equivalent of the AuthTokenFunction in the Find project. This is to allow for development to begin on the standalone auth app.

Unit tests have been added for the new API controller, and the E2E tests have been updated to use the Auth Emulator for local development and ephemeral workflow runs.

## Changes

- Auth Emulator solution and projects created, with functional equivalence with AuthTokenFunction.cs (updated for differences in a non-Azure-Functions application)
- E2E Tests updated to use the Auth Emulator according to a config (true by default)
- Github workflows updated to build new components and pass through new configs

## Validation

- Unit / Integration / E2E tests run locally
- E2E tests passing ephemerally, confirmed it uses the auth emulator endpoint
- E2E tests passing on dev environments, confirmed it does not use the auth emulator endpoint


Apps/AuthEmulator/src/SUI.AuthEmulator/Data/auth-clients-inbound.json
robshakespeare, line 1: **todo:** Please can we not create duplicates of this file.

Please can you do the following:
* Delete this file (`rm Apps/AuthEmulator/src/SUI.AuthEmulator/Data/auth-clients-inbound.json`)
* Update the Auth Emulator's csproj file to include:
```
  <ItemGroup>
    <None Include="..\..\..\..\Data\auth-clients-inbound.json">
      <Link>Data\auth-clients-inbound.json</Link>
      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
    </None>
  </ItemGroup>
```

Thank you

Apps/AuthEmulator/src/SUI.AuthEmulator/SUI.AuthEmulator.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Asp.Versioning.Http" Version="10.0.0" />
+    <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="8.16.0" />
+    <PackageReference Include="Testably.Abstractions.FileSystem.Interface" Version="10.0.0" />
+  </ItemGroup>
+  <ItemGroup>
+    <Content Update="Data\auth-clients-inbound.json">
```
robshakespeare: **todo:** Relating to the other comment about `auth-clients-inbound.json`, please can you replace this section with:

```
  <ItemGroup>
    <None Include="..\..\..\..\Data\auth-clients-inbound.json">
      <Link>Data\auth-clients-inbound.json</Link>
      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
    </None>
  </ItemGroup>
```

Apps/AuthEmulator/global.json
robshakespeare, line 1: **todo:** Please can you delete this file.  I realise it was probably created by a template, but in this monorepo setup we should allow each project to fallback to the `gloabl.json` in the root of the repo, so that we can upgrade .NET versions across the board for easily.  Thanks.

Apps/Find/tests/SUI.Find.E2ETests/Config.cs
```diff
@@ -21,4 +21,8 @@ public record Config
     public DateTimeOffset? CheckFindApiBuildTimestampThreshold { get; init; }
 
     public DateTimeOffset? CheckStubCustodiansApiBuildTimestampThreshold { get; init; }
+
+    public string AuthEmulatorBaseUrl { get; init; } = "https://localhost:7250/api/";
```
robshakespeare: **todo:** Please can this be changed.  Looking at how all this come together, it makes sense to just make the whole token URL configurable rather than have the `UseAuthEmulator` conditional logic.  This will help with the next steps in the auth evolution, including deployment and then eventually using the e2e tests against deployed environments which use real identity providers.

Please can you do the following:
* Change the property here to be `AccessTokenUrl`
* Remove the `UseAuthEmulator`
* Update the e2e workflow accordingly, to specify the full, absolute URL to use (i.e. scheme and domain name) so that ephemeral uses the new auth emulator, but deployed stays using the old auth token URL for now.
* Remove `AuthEmulatorClient`, and just use the existing `FunctionTestFixture.Client` property.
* Pass the full `Config.AccessTokenUrl` value when building the request in `GetAuthTokenWithRetryAsync`

I hope that makes sense?  I'm happy to have a quick call to talk this through 🙂

robshakespeare (changes requested):
Thank you for making these changes so far.

Please can you do the follow-on changes as per my comments.

I'm happy to have a quick call to talk this through? 🙂

robshakespeare (approved):
Brilliant, thank you :+1:

Apps/AuthEmulator/src/SUI.AuthEmulator/Data/auth-clients-inbound.json
robshakespeare: Thanks

Apps/AuthEmulator/global.json
robshakespeare: Thanks

Apps/Find/tests/SUI.Find.E2ETests/Config.cs
robshakespeare: Thank you for making these changes

PR #640, opened 2026-05-28
feat(auth): SUI-1757 - Auth Emulator app: add health check endpoint

jordanstarksbrowning-hippo:
## Summary

Add a health check endpoint for the Auth Emulator app, and update the E2E tests to also check for this endpoint.

## Changes
- Add new controller with health check endpoint
- Add new method in E2E test fixture to include new endpoint in the health check step
- Add new config to isolate the health check endpoint to specific environments

## Validation
- Unit, integration, and E2E tests all pass

### DIFF
--- a/.github/workflows/find-e2e-tests.yml
+++ b/.github/workflows/find-e2e-tests.yml
@@ -236,7 +236,7 @@ jobs:
             --results-directory TestResults \
             -e E2E__BaseUrl=https://${{ steps.env_config.outputs.subscription_prefix }}${{ steps.env_config.outputs.environment_id }}func-${{ steps.env_config.outputs.region_short }}-find01.azurewebsites.net/api/ \
             -e E2E__StubCustodiansBaseUrl=https://${{ steps.env_config.outputs.subscription_prefix }}${{ steps.env_config.outputs.environment_id }}app-${{ steps.env_config.outputs.region_short }}-custodians01.azurewebsites.net/api/ \
-            -e E2E__AuthEmulatorHealthCheckEndpoint=null \
+            -e E2E__AuthEmulatorHealthCheckEndpoint="" \
             -e E2E__SkipResetAzureTables=True \
             -e E2E__AccessTokenUrl="v1/auth/token" \
             -e E2E__CheckStubCustodiansApiBuildTimestampThreshold=${{ contains(github.event.workflow_run.name, 'Custodian') && github.event.workflow_run.run_started_at || '' }} \
--- a/.github/workflows/find-e2e-tests.yml
+++ b/.github/workflows/find-e2e-tests.yml
@@ -189,6 +189,7 @@ jobs:
             -e E2E__BaseUrl=http://localhost:7182/api/ \
             -e E2E__StubCustodiansBaseUrl=http://localhost:5193/api/ \
             -e E2E__AccessTokenUrl=http://localhost:5250/api/v1/auth/token \
+            -e E2E__AuthEmulatorHealthCheckEndpoint=http://localhost:5250/api/health \
             -e E2E__SkipResetAzureTables=False \
             -e E2E__CheckStubCustodiansApiBuildTimestampThreshold=${{ needs.eval-workflow-setup.outputs.job_start_time }} \
             -e E2E__CheckFindApiBuildTimestampThreshold=${{ needs.eval-workflow-setup.outputs.job_start_time }} # The threshold checks are not necessary for ephemeral e2e tests, but are included to verify the threshold behaviour at pull request stage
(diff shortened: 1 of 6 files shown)


.github/workflows/find-e2e-tests.yml
```diff
@@ -235,6 +236,7 @@ jobs:
             --results-directory TestResults \
             -e E2E__BaseUrl=https://${{ steps.env_config.outputs.subscription_prefix }}${{ steps.env_config.outputs.environment_id }}func-${{ steps.env_config.outputs.region_short }}-find01.azurewebsites.net/api/ \
             -e E2E__StubCustodiansBaseUrl=https://${{ steps.env_config.outputs.subscription_prefix }}${{ steps.env_config.outputs.environment_id }}app-${{ steps.env_config.outputs.region_short }}-custodians01.azurewebsites.net/api/ \
+            -e E2E__AuthEmulatorHealthCheckEndpoint=null \
```
robshakespeare: **todo:** This just passes the text value of "null".  In bash I'm not there's a way to pass null, and we should just pass any empty string:
```
-e E2E__AuthEmulatorHealthCheckEndpoint= \
```
And then we need to check for it being **not null or empty** in the E2E test logic.

(I checked the logs of the d01 e2e test run, and it does indeed try and check this: https://s270d01func-ukw-find01.azurewebsites.net/api/null)

Apps/Find/tests/SUI.Find.E2ETests/FunctionTestFixture.cs
```diff
@@ -64,21 +65,39 @@ public async Task EnsureFindApiIsUpAsync(ITestOutputHelper testOutputHelper)
...
         await EnsureServiceIsUpAsync(
             "StubCustodians API",
             StubCustodiansClient,
+            HealthEndpointUrl,
             testOutputHelper,
             checkBuildTimestampThreshold: Config.CheckStubCustodiansApiBuildTimestampThreshold
         );
     }
 
+    private async Task EnsureAuthEndpointApiIsUpAsync(ITestOutputHelper testOutputHelper)
```
robshakespeare: **nitpick:** Please can you rename this to `EnsureAuthEmulatorApiIsUpAsync`

Apps/Find/tests/SUI.Find.E2ETests/FunctionTestFixture.cs
```diff
@@ -64,21 +65,39 @@ public async Task EnsureFindApiIsUpAsync(ITestOutputHelper testOutputHelper)
...
+        if (!string.IsNullOrEmpty(Config.AuthEmulatorHealthCheckEndpoint))
+        {
+            TestContext.Current.SendDiagnosticMessage("Checking Auth Emulator API health...");
+
+            await EnsureServiceIsUpAsync(
+                "AuthEmulator API",
+                Client,
+                Config.AuthEmulatorHealthCheckEndpoint,
+                testOutputHelper,
+                checkBuildTimestampThreshold: Config.CheckFindApiBuildTimestampThreshold
```
robshakespeare: **todo:** Please can you change this to be:

```
Config.CheckAuthEmulatorApiBuildTimestampThreshold
```

It will cause false failures if use the FindApi's build timestamp, because it most cases it will wait forever because Find will be deployed more frequently than the emulator.

The config value can be null, and we can just leave it as null for now.  We'd only need to pass an actual value if start to get the timeout issues we saw with the other apps.

Apps/Find/tests/SUI.Find.E2ETests/FunctionTestFixture.cs
```diff
@@ -152,11 +172,11 @@ private async Task EnsureTablesResetOneTimeInitAsync(ITestOutputHelper testOutpu
     private static async Task EnsureServiceIsUpAsync(
         string serviceName,
         HttpClient client,
+        string url,
```
robshakespeare, line 175: **question:** Is this change needed now, given it will only ever be checking out `health` endpoints?

.github/workflows/find-e2e-tests.yml
robshakespeare: ah ha, beat me to it! 🙂

Apps/Find/tests/SUI.Find.E2ETests/FunctionTestFixture.cs
jordanstarksbrowning-hippo: So I've done this change to allow for the Auth Emulator method to use the absolute health check endpoint URL with an existing client, it only uses "health" for the find and stubs apis

robshakespeare: ah yes, of course!  Thanks, yes I understand why its this way now.

Apps/Find/tests/SUI.Find.E2ETests/FunctionTestFixture.cs
robshakespeare: Thanks for making the change :+1:

Apps/Find/tests/SUI.Find.E2ETests/FunctionTestFixture.cs
robshakespeare: Thanks for making the change :+1:

robshakespeare (approved):
Thank you for adding this health check logic, and making the changes

PR #642, opened 2026-05-28
chore: remove invalid encryption property from env variables

udara-jay:


robshakespeare (approved):
Thanks for the diligence on removing all of the Person ID encryption stuff :+1:

PR #648, opened 2026-06-01
feat(AuthEmulator): SUI-1776 - Add JWKS endpoint to Auth Emulator

udara-jay:
## Summary

Add a JWKS endpoint to the Auth Emulator, so that we avoid auth bypasses becoming part of the architecture (i.e. Local / CI / Ephemeral / Dev environments should still provide and use OIDC).

## Changes

- Class-level service(s):
> Class-level service that generates 4 RSA Keys in the constructor, and stores them in a private field
> Registered as a singleton

- JSON Web Key Set (JWKS) endpoint:
> Endpoint is mapped to GET api/v1/jwks (in a new JWKSController)
> Endpoint should return the public key information in the format as described below.

## Validation

Successful unit test runs.

### DIFF
--- a/Apps/AuthEmulator/src/SUI.AuthEmulator/Controllers/JWKSController.cs
+++ b/Apps/AuthEmulator/src/SUI.AuthEmulator/Controllers/JWKSController.cs
@@ -1,0 +1,38 @@
+using Microsoft.AspNetCore.Mvc;
+using SUI.AuthEmulator.Services;
+
+namespace SUI.AuthEmulator.Controllers;
+
+[ApiController]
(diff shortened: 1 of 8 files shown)


robshakespeare (dismissed):
This looks good to me, thanks :+1:

I noticed that the GitHub Code Quality bot has spotted some issues with not disposing the RSA keys.  However, because we know they are singletons that live for the entire lifetime of the application, and because this is a mock, I am fine with dismissing the CodeQL issues.  It would add unnecessary complexity to this app otherwise.

robshakespeare (approved):
Looks good, thanks 👍

PR #649, opened 2026-06-01
chore: SUI-1794 - Tidy the generated OpenAPI spec for the find (search) and search results endpoints

robshakespeare:
## Summary

This PR improves and fixes the generated OpenAPI spec for the find (search) and search results endpoints; so that we move the generated spec closer to one that suppliers can integrate with, and so that this part of the spec is ready to be shared with the wider programme of work.

## Changes

- Corrected the name of the search endpoint's response model from `SearchJobV2` to `SearchWorkItem`.
- Added missing 401 and 500 responses.
- Improved and tidied descriptions.
- Improved and fixed the tracing header names in the OpenAPI spec.
- Resolved the enums being incorrectly `int` rather than `string` in Find’s generated OpenAPI spec.
    - Generic implementation is a workaround because Azure Function’s do not have the Schema Transformer ability.
- The payload property of `CustodianRecord` was incorrectly getting generated in the OpenAPI spec as a schema of type `JsonElement`.  This PR fixes that to be an open object whose schema will match the `SchemaUri`, which best represents the intention of the API.
- Refactored `FindDocumentFilter.Apply()` method in to smaller methods so its easier to read and follow.

## Validation

- Generating the OpenAPI spec and manual verification it.

### DIFF
--- a/Apps/Find/src/SUI.Find.Application/Models/CustodianRecord.cs
+++ b/Apps/Find/src/SUI.Find.Application/Models/CustodianRecord.cs
@@ -15,7 +15,7 @@ public record CustodianRecord
 
     [OpenApiProperty(
         Nullable = true,
-        Description = "The payload of the record, if available. An open object that can contain any number and type of properties, as defined by the `RecordType` and `SchemaUri` properties of this object."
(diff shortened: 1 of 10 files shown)


udara-jay (approved):
Looks great! Thanks for cleaning things up a bit more 👍

PR #650, opened 2026-06-02
feat(authEmulator): SUI-1777 - Add OIDC discovery endpoint to Auth Emulator

udara-jay:
## Summary

Add an OIDC discovery endpoint to the Auth Emulator, so that we avoid auth bypasses becoming part of the architecture (i.e. Local / CI / Ephemeral / Dev environments should still provide and use OIDC).

## Changes

- Endpoint is mapped to GET api/v1/.well-known/openid-configuration (in a new OIDCDiscoveryController).
- The config values are loaded via a .NET Options Pattern object called AuthSettings.
- The BaseUrl for the deployed environments should be specified via App Settings in the Terraform:
> AuthSettings__BaseUrl with value of format("https://%s%sapp-%s-authemulator01.azurewebsites.net/", var.subscription_prefix, var.environment_id, var.region_short)

## Validation

Successful Unit tests / terraform plan and application

### DIFF
--- a/Apps/AuthEmulator/src/SUI.AuthEmulator/Controllers/OIDCDiscoveryController.cs
+++ b/Apps/AuthEmulator/src/SUI.AuthEmulator/Controllers/OIDCDiscoveryController.cs
@@ -1,0 +1,37 @@
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Options;
+using SUI.AuthEmulator.Configurations;
+
+namespace SUI.AuthEmulator.Controllers;
+
+[ApiController]
+[Route("api/v1/.well-known/openid-configuration")]
+public class OIDCDiscoveryController : ControllerBase
+{
+    private readonly AuthSettings _authSettings;
+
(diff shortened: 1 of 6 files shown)


Apps/AuthEmulator/src/SUI.AuthEmulator/Controllers/OIDCDiscoveryController.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    public OIDCDiscoveryController(IOptions<AuthSettings> authSettings)
+    {
+        _authSettings = authSettings.Value;
+    }
+
+    [HttpGet]
+    public IActionResult GetDiscoveryDocument()
+    {
+        var baseUrl = _authSettings.BaseUrl.TrimEnd('/');
```
robshakespeare, line 21: **praise:** Nice, good thinking with sanitising the base URL (truncation logic) :+1:

robshakespeare (approved):
Super, thank you

PR #656, opened 2026-06-04
Update Auth Middleware to Map ClientId to OrganisationId

jordanstarksbrowning-hippo:
## Summary

We want to eventually move to using the organisationId instead of the clientId in the auth mechanism in Find, because this is the ID we will have control over.

This PR takes a step in that direction. We have added OrganisationId to the AuditContext object, and replaced each usage of ClientId to use OrganisationId instead.

We have also added unit tests to the JwtAuthMiddleware class so that the more complex functionality can be covered.

## Changes

- Add OrganisationId to AuthContext
- Replace all usages of AuthContext.ClientId to AuthContext.OrganisationId
- Add unit tests to JwtAuthMiddleware, and refactor to allow testability

## Validation

- New unit tests added

### DIFF
--- a/Apps/Find/src/SUI.Find.Application/Services/MatchPersonOrchestrationService.cs
+++ b/Apps/Find/src/SUI.Find.Application/Services/MatchPersonOrchestrationService.cs
@@ -17,7 +17,7 @@ public class MatchPersonOrchestrationService(
 {
     public async Task<OneOf<string, DataQualityResult, NotFound, Error>> FindPersonIdAsync(
         PersonSpecification specification,
-        string clientId,
+        string organisationId,
         CancellationToken ct
     )
     {
@@ -25,7 +25,10 @@ public class MatchPersonOrchestrationService(
 
         if (matchResult.TryPickT0(out var personId, out var remainder))
         {
-            var getPersonIdResult = await HandlePersonIdRepresentationAsync(personId, clientId);
+            var getPersonIdResult = await HandlePersonIdRepresentationAsync(
+                personId,
+                organisationId
+            );
             return getPersonIdResult.Match<OneOf<string, DataQualityResult, NotFound, Error>>(
                 plainId => plainId,
                 error => error
@@ -41,13 +44,16 @@ public class MatchPersonOrchestrationService(
 
     private async Task<OneOf<string, Error>> HandlePersonIdRepresentationAsync(
         NhsPersonId nhsPersonId,
-        string clientId
+        string organisationId
     )
     {
-        var client = await custodianService.GetCustodianAsync(clientId);
-        if (!client.Success || client.Value is null)
+        var organisation = await custodianService.GetCustodianAsync(organisationId);
+        if (!organisation.Success || organisation.Value is null)
         {
-            logger.LogError("Custodian organisation not found for client ID: {ClientId}", clientId);
+            logger.LogError(
+                "Custodian organisation not found for organisation ID: {OrganisationId}",
+                organisationId
+            );
             return new Error();
         }
 
--- a/Apps/Find/src/SUI.Find.Application/Services/SearchService.cs
+++ b/Apps/Find/src/SUI.Find.Application/Services/SearchService.cs
@@ -18,7 +18,7 @@ public interface ISearchService
 {
     Task<OneOf<SearchJobDto, Error>> StartSearchAsync(
         string inputPersonId,
-        string clientId,
+        string organisationId,
         DurableTaskClient client,
         string correlationId,
         CancellationToken cancellationToken
@@ -26,21 +26,21 @@ public interface ISearchService
 
     Task<OneOf<SearchJobDto, NotFound, Forbidden, Error>> CancelSearchAsync(
         string jobId,
-        string clientId,
+        string organisationId,
         DurableTaskClient client,
         CancellationToken cancellationToken
     );
 
     Task<OneOf<SearchResultsDto, NotFound, Forbidden, Error>> GetSearchResultsAsync(
         string jobId,
-        string clientId,
+        string organisationId,
         DurableTaskClient client,
         CancellationToken cancellationToken
     );
 
     Task<OneOf<SearchJobDto, Forbidden, NotFound, Error>> GetSearchStatusAsync(
         string jobId,
-        string clientId,
+        string organisationId,
         DurableTaskClient client,
         CancellationToken cancellationToken
     );
@@ -56,13 +56,13 @@ public class SearchService(
 {
     public async Task<OneOf<SearchJobDto, Error>> StartSearchAsync(
         string inputPersonId,
-        string clientId,
+        string organisationId,
         DurableTaskClient client,
         string correlationId,
         CancellationToken cancellationToken
     )
     {
-        var instanceId = $"{inputPersonId}-{clientId}";
+        var instanceId = $"{inputPersonId}-{organisationId}";
         var hashedInstanceId = hashService.HmacSha256Hash(instanceId);
 
         var existingInstance = await client.GetInstanceAsync(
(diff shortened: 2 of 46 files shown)


Apps/Find/src/SUI.Find.Application/Services/MatchPersonOrchestrationService.cs
```diff
@@ -41,13 +44,16 @@ CancellationToken ct
...
     )
     {
-        var client = await custodianService.GetCustodianAsync(clientId);
-        if (!client.Success || client.Value is null)
+        var organisation = await custodianService.GetCustodianAsync(organisationId);
+        if (!organisation.Success || organisation.Value is null)
         {
-            logger.LogError("Custodian organisation not found for client ID: {ClientId}", clientId);
+            logger.LogError(
+                "Custodian organisation not found for organisation ID: {OrganisationId}",
```
robshakespeare, line 54: **praise:** Thank you for the attention to detail in this set of changes, including parameter names, variable names and even log statements :+1:

Apps/Find/src/SUI.Find.Application/Services/SearchService.cs
```diff
@@ -164,7 +164,7 @@ CancellationToken cancellationToken
 
             // Check if the job belongs to the requesting client
             var input = ReadOrchestratorInput<SearchOrchestratorInput>(metaData);
-            if (input is null || input.PolicyContext.ClientId != clientId)
+            if (input is null || input.PolicyContext.ClientId != organisationId)
```
robshakespeare: **todo:** Please can you rename `ClientId` on the `PolicyContext` class to be called `OrganisationId`.

I have checked through the usages and it is always referring to an Organisation ID now (and really always was), so `OrganisationId` is the correct name.

Thank you

Apps/Find/tests/SUI.Find.FindApi.UnitTests/MiddlewareTests/JwtAuthMiddlewareTests.cs
```diff
@@ -0,0 +1,295 @@
...
+using SUI.Find.Application.Constants;
+using SUI.Find.FindApi.Middleware;
+using SUI.Find.FindApi.Models;
+using SUI.Find.FindApi.UnitTests.Mocks;
+using SUI.Find.Infrastructure.Models;
+using SUI.Find.Infrastructure.Services;
+
+namespace SUI.Find.FindApi.UnitTests.MiddlewareTests;
+
+public class JwtAuthMiddlewareTests
```
robshakespeare, line 19: **praise:** Thanks so much for writing all of this test coverage 🤩

Apps/Find/src/SUI.Find.FindApi/Models/AuthContext.cs
```diff
@@ -3,4 +3,8 @@
 namespace SUI.Find.FindApi.Models;
 
 [ExcludeFromCodeCoverage(Justification = "Will be tested with integration tests.")]
```
robshakespeare: **todo:** Please can you remove this `ExcludeFromCodeCoverage` attribute now.  It now will be implicitly covered by the unit tests added in this PR, and that is enough coverage for plain records.  Thanks

Apps/Find/src/SUI.Find.FindApi/Middleware/JwtAuthMiddleware.cs
```diff
@@ -111,7 +109,7 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
             return;
         }
 
-        var authContext = AuthContextFactory.FromJwt(jwt);
+        var authContext = authContextFactory.FromJwt(jwt, store);
```
robshakespeare, line 112: **praise:** Elegantly refactored to enable testability, good thinking about passing auth store this way too :+1:

Apps/Find/tests/SUI.Find.Application.UnitTests/Services/PersonIdRepresentationServiceTests/FindPersonIdAsyncTests.cs
```diff
@@ -45,7 +45,7 @@ public async Task ShouldReturnPlainPersonId_WhenMatchingServiceReturnsMatchedNhs
         // Act
         var result = await _sut.FindPersonIdAsync(
             specification: personSpec,
-            clientId: "test-client-id",
+            organisationId: "test-client-id",
```
robshakespeare: **nitpick:** Sorry to be picky, but I think its worth updating the test values, to avoid confusion in the future.

Please can you change `test-client-id` to be `test-org-id` everywhere in this file.

Apps/Find/tests/SUI.Find.FindApi.UnitTests/MiddlewareTests/JwtAuthMiddlewareTests.cs
```diff
@@ -0,0 +1,295 @@
...
+    [InlineData("v1/auth/token")]
+    [InlineData("health")]
+    public async Task TestInvoke_WithNoAuthEndpoints_SkipsMethod(string endpoint)
+    {
+        var sut = new JwtAuthMiddleware(_authStoreService, _authContextFactory, _handler);
+        var request = Substitute.For<HttpRequestData>(_context);
+        request.Url.Returns(new Uri("/api/" + endpoint));
+        _context.GetHttpRequestDataAsync().Returns(request);
+
+        await sut.Invoke(_context, Next);
```
robshakespeare, line 41: **suggestion:** I'm a big fan of having `ACT` and `ASSERT` comments in tests, so its easy to scan the tests to see where the 'action under test' takes place, and where the assertions begin.  It is a pattern that is mostly followed in this codebase, and I'd like to encourage that pattern.

If its not too much trouble, it would be great to update the tests in this PR to follow that pattern.  I'll mention it in planning too so the whole team is onboard.  Thanks

robshakespeare (changes requested):
Absolutely excellent set of changes here.  Thanks for doing this so thoroughly, and many thanks for adding the test coverage too.

Just one suggestion and a couple of todo's, thanks.

jordanstarksbrowning-hippo:
Thanks for the suggestions Rob, and appreciate the comments here. While implementing them, I found a few more places with client Id leftover, so I changed those too.

Apps/Find/src/SUI.Find.Application/Services/SearchService.cs
robshakespeare: Thanks

Apps/Find/src/SUI.Find.FindApi/Models/AuthContext.cs
robshakespeare: Thanks

Apps/Find/tests/SUI.Find.Application.UnitTests/Services/PersonIdRepresentationServiceTests/FindPersonIdAsyncTests.cs
robshakespeare: Thanks :+1:

Apps/Find/tests/SUI.Find.FindApi.UnitTests/MiddlewareTests/JwtAuthMiddlewareTests.cs
robshakespeare: Thanks for making the suggested changes

robshakespeare:
> Thanks for the suggestions Rob, and appreciate the comments here. While implementing them, I found a few more places with client Id leftover, so I changed those too.

Thanks Jordan.  Good spots with the other client id places too :+1:

robshakespeare (approved):
Superb, thanks

Thank you for making the changes, and good spot with the "Ord" -> "Org", and the personId.Value bits.  Great updates.

PR #657, opened 2026-06-05
feat(find): SUI-1800 - Include ClientId when auditing access to HTTP endpoints

jordanstarksbrowning-hippo:
## Summary

As part of the work to separate Client IDs from Organisation IDs, Client IDs were replaced by Organisation IDs in the auditing process. 

This PR is to include the Client IDs specifically when auditing access to HTTP endpoints, so that we can still build a clear picture of HTTP access events.

## Changes

- Add Client ID to AuditAccessMessage.cs
- Fill this value with the Client ID from the audit context in AuditMiddleware.cs
- Update tests
- Fix another missing ClientId/OrgId swap

## Validation


robshakespeare (approved):
Thanks :+1:

PR #661, opened 2026-06-08
feat(find): sui-1778 - Add OIDC Discovery and RSA signature verification in Find

udara-jay:
## Summary

Add OIDC Discovery and RSA signature verification to the Find API’s Authentication functionality, so that we can support integrating with common 3rd party identity providers.

## Changes

- OpenID Discovery included using the Microsoft.IdentityModel.Protocols.OpenIdConnect package
> OIDC Configuration Manager is registered in host services as a singleton, so it caches the public keys (JWKS) across function invocations

- Verify token signatures using RSA public key
> Use Microsoft.IdentityModel.Protocols.IConfigurationManager<OpenIdConnectConfiguration>
> Call GetConfiguration in every request (the .NET package deals with caching and expiry) var oidcConfig = await oidcConfigManager.GetConfigurationAsync(cancellationToken);
> Call JwtSecurityTokenHandler.ValidateToken and pass the specified Token Validation Parameters:
> The config values are loaded via a .NET Options Pattern object called AuthSettings

- Token validation handles keys not being found (edge case that the identity provider rotates keys within a cache window) - see attachment
> When the token validation throws a SecurityTokenSignatureKeyNotFoundException
> And the unverified token’s issuer matches the configured issuer
> And the unverified token’s audiences contains the configured audience
> And this retry logic has not yet been executed in this request
> Then the OIDC configuration manager should be forced refreshed: oidcConfigManager.RequestRefresh();
> And the token validation logic should be retried (suggestion is to a have a singularly recursive method, something like ValidateAsymmetricToken(bool retry) where retry is true first, and then false the second time)

- JwtAuthMiddleware for the general and asymmetric token validation logic covered with unit tests

- The existing Symmetric Key verification is kept and remains unchanged
> Which IssuerSigningKey gets used should be derived from the token’s alg header
> RS256 should use the new RsaSecurityKey verification
> HS256 should use the existing SymmetricSecurityKey verification
> Symmetric Key verification will be deleted at a later stage to avoid any breaking changes

## Validation

- Scenario: Valid token, should Allow access, and the auth context should be set as expected
> Given a valid token
> When the token is verified
> Then the result should Allow access
> And there should be no problem response
> And the auth context should be for the expected ClientId, OrganisationId and have the expected Scopes

- Scenario: Valid token, except for kid is missing in our OIDC Config, so force refresh should happen, and the final result should Allow access, and the auth context should be set as expected
> Given a valid token
> And the token’s public key is no longer available in the oidcConfig
> When the token is verified
> Then a refresh should happen
> And the token’s public key becomes available in the oidcConfig
> Then the result should Allow access
> And there should be no problem response

- Scenario: Valid token, except for kid is missing in our oidcConfig, so force refresh should happen, and after force refresh the kid is still missing, then should Deny access
> Given a valid token
> And the token’s public key is no longer available in the oidcConfig
> When the token is verified
> Then a refresh should happen
> And the token’s public key is still not available in the oidcConfig
> Then the result should Deny access with a problem response
> And the problem response should indicate the signature’s key was not found

- Scenario: Invalid signature due to modified token payload, should Deny access
> Given a valid token
> And the token's payload is changed
> When the token is verified
> Then the result should Deny access with a problem response
> And the problem response should indicate an invalid signature

- Scenario: Invalid signature due to modified token header, should Deny access
> Given a valid token
> And the token's header is changed
> When the token is verified
> Then the result should Deny access with a problem response
> And the problem response should indicate an invalid signature

- Scenario: Invalid due to use of token before valid time range (i.e. not-before), should Deny access
> Given a token that is valid between 2 hours and 3 hours
> When the token is verified
> Then the result should Deny access with a problem response
> And the problem response should indicate the jwt is not active

- Scenario: Invalid due to use of token after valid time range (i.e. expiration), should Deny access
> Given a token that is valid between -3 hours and -2 hours
> When the token is verified
> Then the result should Deny access with a problem response
> And the problem response should indicate the jwt is expired

- Scenario: Invalid signature due to signature removal, should Deny access
> Given a valid token
> And the token's signature is removed
> When the token is verified
> Then the result should Deny access with a problem response
> And the problem response should indicate the jwt signature is required

- Scenario: Invalid signature due to non-genuine private key, should Deny access
> Given a valid token except for being signed with a different RSA private key
> When the token is verified
> Then the result should Deny access with a problem response
> And the problem response should indicate an invalid signature

- Scenario: Invalid source (i.e. issuer), should Deny access
> Given a valid token but from a different issuer
> When the token is verified
> Then the result should Deny access with a problem response
> And the problem response should indicate that the jwt issuer is invalid

- Scenario: Invalid target (i.e. audience), should Deny access
> Given a valid token but intended for different audience
> When the token is verified
> Then the result should Deny access with a problem response
> And the problem response should indicate that the jwt audience is invalid

### DIFF
--- a/Apps/Find/src/SUI.Find.FindApi/Configurations/AuthSettings.cs
+++ b/Apps/Find/src/SUI.Find.FindApi/Configurations/AuthSettings.cs
@@ -1,0 +1,13 @@
+namespace SUI.Find.FindApi.Configurations;
+
+public class AuthSettings
+{
+    public const string SectionName = "AuthSettings";
+
+    // OIDC Discovery configuration properties
+    public string Issuer { get; set; } =
+        "https://sandbox.api.example.gov.uk/sui-find-a-record/auth";
+    public string Audience { get; set; } = "sui-find-a-record-api";
+    public string OidcDiscoveryUrl { get; set; } =
+        "https://sandbox.api.example.gov.uk/sui-find-a-record/auth/.well-known/openid-configuration";
+}
--- a/Apps/Find/src/SUI.Find.FindApi/Configurations/AuthSettings.cs
+++ b/Apps/Find/src/SUI.Find.FindApi/Configurations/AuthSettings.cs
@@ -9,5 +9,5 @@ public class AuthSettings
         "https://sandbox.api.example.gov.uk/sui-find-a-record/auth";
     public string Audience { get; set; } = "sui-find-a-record-api";
     public string OidcDiscoveryUrl { get; set; } =
-        "https://sandbox.api.example.gov.uk/sui-find-a-record/auth/.well-known/openid-configuration";
+        "https://localhost:7250/api/v1/.well-known/openid-configuration";
 }
--- a/Apps/Find/src/SUI.Find.FindApi/Middleware/JwtAuthMiddleware.cs
+++ b/Apps/Find/src/SUI.Find.FindApi/Middleware/JwtAuthMiddleware.cs
@@ -4,6 +4,7 @@ using System.Reflection;
 using System.Text;
 using Microsoft.Azure.Functions.Worker;
 using Microsoft.Azure.Functions.Worker.Middleware;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.IdentityModel.Protocols;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
@@ -21,7 +22,8 @@ public class JwtAuthMiddleware(
     IAuthStoreService authStoreService,
     IAuthContextFactory authContextFactory,
     IConfigurationManager<OpenIdConnectConfiguration> oidcConfigManager,
-    IOptions<AuthSettings> authSettings
+    IOptions<AuthSettings> authSettings,
+    ILogger<JwtAuthMiddleware> logger
 ) : IFunctionsWorkerMiddleware
 {
     // One concrete handler with no interfaces, no injection mocks.
@@ -125,16 +127,18 @@ public class JwtAuthMiddleware(
         }
         catch (SecurityTokenException ex)
         {
+            logger.LogWarning(ex, "Token validation failed");
             context.GetInvocationResult().Value = await HttpResponseUtility.ProblemResponse(
                 req,
                 HttpStatusCode.Unauthorized,
                 nameof(HttpStatusCode.Unauthorized),
-                ex.Message
+                "Token validation failed."
             );
             return;
         }
-        catch
+        catch (Exception ex)
         {
+            logger.LogWarning(ex, "Invalid bearer token");
             context.GetInvocationResult().Value = await HttpResponseUtility.ProblemResponse(
                 req,
                 HttpStatusCode.Unauthorized,
--- a/Apps/Find/src/SUI.Find.FindApi/Middleware/JwtAuthMiddleware.cs
+++ b/Apps/Find/src/SUI.Find.FindApi/Middleware/JwtAuthMiddleware.cs
@@ -4,22 +4,29 @@ using System.Reflection;
 using System.Text;
 using Microsoft.Azure.Functions.Worker;
 using Microsoft.Azure.Functions.Worker.Middleware;
+using Microsoft.Extensions.Options;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 using Microsoft.IdentityModel.Tokens;
 using SUI.Find.Application.Constants;
 using SUI.Find.FindApi.Attributes;
+using SUI.Find.FindApi.Configurations;
 using SUI.Find.FindApi.Models;
 using SUI.Find.FindApi.Utility;
 using SUI.Find.Infrastructure.Services;
 
 namespace SUI.Find.FindApi.Middleware;
 
-// ReSharper disable once ClassNeverInstantiated.Global
 public class JwtAuthMiddleware(
     IAuthStoreService authStoreService,
     IAuthContextFactory authContextFactory,
-    ISecurityTokenValidator handler
+    IConfigurationManager<OpenIdConnectConfiguration> oidcConfigManager,
+    IOptions<AuthSettings> authSettings
 ) : IFunctionsWorkerMiddleware
 {
+    // One concrete handler with no interfaces, no injection mocks.
+    private static readonly JwtSecurityTokenHandler TokenHandler = new();
+
     public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
     {
         var req = await context.GetHttpRequestDataAsync();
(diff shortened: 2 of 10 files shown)


Apps/Find/src/SUI.Find.FindApi/Program.cs
```diff
@@ -44,6 +50,17 @@
 // .NET services
 builder.Services.AddSingleton(TimeProvider.System);
 
+// Register ConfigurationManager as a Singleton to cache public keys across function invocations
```
robshakespeare: **nitpick:** Please can OIDC be included in the this comment, to clarify what the ConfigurationManager is for.

```suggestion
// Register OpenID Connect ConfigurationManager as a Singleton to cache public keys across function invocations
```

(I realise this probably originated from something I wrote, but seeing it in this context made me realise being more specific will help 🙂 )

Apps/Find/src/SUI.Find.FindApi/Configurations/AuthSettings.cs
```diff
@@ -0,0 +1,13 @@
...
+public class AuthSettings
+{
+    public const string SectionName = "AuthSettings";
+
+    // OIDC Discovery configuration properties
+    public string Issuer { get; set; } =
+        "https://sandbox.api.example.gov.uk/sui-find-a-record/auth";
+    public string Audience { get; set; } = "sui-find-a-record-api";
+    public string OidcDiscoveryUrl { get; set; } =
+        "https://sandbox.api.example.gov.uk/sui-find-a-record/auth/.well-known/openid-configuration";
```
robshakespeare: **todo:** I think even though this is a more sensible URL, it would be good for the default URL to actually work for local dev (because of the existing issue of having to copy config from `example.local.settings.json`).


```suggestion
        "https://localhost:7250/api/v1/.well-known/openid-configuration";
```

Apps/Find/src/SUI.Find.FindApi/Middleware/JwtAuthMiddleware.cs
```diff
@@ -129,10 +164,57 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+        {
+            RequireSignedTokens = true,
+            ValidateIssuer = true,
+            ValidIssuer = configValues.Issuer,
+            ValidateAudience = true,
+            ValidAudience = configValues.Audience,
+            ValidateIssuerSigningKey = true,
+            IssuerSigningKeys = oidcConfig.SigningKeys,
+            ValidateLifetime = true,
+            ClockSkew = TimeSpan.FromMinutes(2),
```
robshakespeare: **todo:** Please can this be removed.  I realise the old implementation specified clock skew of 2 minutes, but the Microsoft library's default is 5 minutes and I think we should use that as the default.  Thank you 🙂 

We can just remove the whole line, and then the default will be used:
```suggestion
```

terraform/environments/d01.tfvars
```diff
@@ -23,9 +22,14 @@ find_app_settings = {
   NhsAuthConfig__NHS_DIGITAL_TOKEN_URL                       = "https://int.api.service.nhs.uk/oauth2/token"
   NhsAuthConfig__NHS_DIGITAL_FHIR_ENDPOINT                   = "https://int.api.service.nhs.uk/personal-demographics/FHIR/R4/"
   NhsAuthConfig__NHS_DIGITAL_ACCESS_TOKEN_EXPIRES_IN_MINUTES = 5
+  AuthSettings__Issuer           = "https://sandbox.api.example.gov.uk/sui-find-a-record/auth"
+  AuthSettings__Audience         = "sui-find-a-record-api"
+  AuthSettings__OidcDiscoveryUrl = "https://sandbox.api.example.gov.uk/sui-find-a-record/auth/.well-known/openid-configuration"
```
robshakespeare: **issue:** Great that you included this, I think I had overlooked it!  But, this value needs to actually point to the real endpoint:
 
```suggestion
  AuthSettings__OidcDiscoveryUrl = "https://s270d01app-ukw-authemulator01.azurewebsites.net/api/v1/.well-known/openid-configuration"
```

Please can you update this here, and for d02.

Also, please could you apply the update for d01 only?

terraform/environments/d02.tfvars
```diff
@@ -23,12 +22,17 @@ find_app_settings = {
   NhsAuthConfig__NHS_DIGITAL_TOKEN_URL                       = "https://int.api.service.nhs.uk/oauth2/token"
   NhsAuthConfig__NHS_DIGITAL_FHIR_ENDPOINT                   = "https://int.api.service.nhs.uk/personal-demographics/FHIR/R4/"
   NhsAuthConfig__NHS_DIGITAL_ACCESS_TOKEN_EXPIRES_IN_MINUTES = 5
+  AuthSettings__Issuer           = "https://sandbox.api.example.gov.uk/sui-find-a-record/auth"
+  AuthSettings__Audience         = "sui-find-a-record-api"
+  AuthSettings__OidcDiscoveryUrl = "https://sandbox.api.example.gov.uk/sui-find-a-record/auth/.well-known/openid-configuration"
```
robshakespeare: **issue:** Please can this point to the real endpoint:
 
```suggestion
  AuthSettings__OidcDiscoveryUrl = "https://s270d02app-ukw-authemulator01.azurewebsites.net/api/v1/.well-known/openid-configuration"
```

Apps/Find/src/SUI.Find.FindApi/Middleware/JwtAuthMiddleware.cs
```diff
@@ -169,10 +251,5 @@ private static string[] GetRequiredScopes(FunctionContext context)
     private static bool HasAnyRequiredScope(
         AuthContext caller,
         IReadOnlyList<string> requiredScopes
-    )
-    {
-        return requiredScopes.Any(rs =>
-            caller.Scopes.Contains(rs, StringComparer.OrdinalIgnoreCase)
-        );
-    }
+    ) => requiredScopes.Any(rs => caller.Scopes.Contains(rs, StringComparer.OrdinalIgnoreCase));
```
robshakespeare: **nitpick:** I don't think this actually changed any functionality did it?  It was just a code style change.

Ordinarily I would say this was a good thing, but in this case Jordan is currently working in this specific area (scopes checking), so I think its best we leave it alone to help avoid merge conflicts.

Please you revert the change here (`HasAnyRequiredScope`) if it was just a code style change? 🙂  Thank you

Apps/Find/tests/SUI.Find.FindApi.UnitTests/MiddlewareTests/JwtAuthMiddlewareTests.cs
```diff
@@ -1,313 +1,745 @@
...
+        if (authHeaderValue != null)
+            requestHeaders.Add("Authorization", authHeaderValue);
+        requestData.Headers.Returns(requestHeaders);
+
+        var responseHeaders = new HttpHeadersCollection();
+        responseData.Headers.Returns(responseHeaders);
+
+        requestData.Url.Returns(new Uri("https://mock.gov.uk/api/v1/searches"));
+        requestData.CreateResponse().Returns(responseData);
+        responseData.Body.Returns(new MemoryStream());
```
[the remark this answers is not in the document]
robshakespeare: This makes sense to do, and I have tried it locally.  I'll commit the suggestion.

Apps/Find/tests/SUI.Find.FindApi.UnitTests/MiddlewareTests/JwtAuthMiddlewareTests.cs
```diff
@@ -1,313 +1,745 @@
...
+            );
+
+            // Assert
+            Assert.True(nextExecuted);
+            Assert.Null(context.GetInvocationResult().Value);
+            Assert.Equal(authContext, context.Items[ApplicationConstants.Auth.AuthContextKey]);
+        }
+
+        [Fact]
+        public async Task Scenario2_KidMissingInitially_ShouldForceRefreshAndSucceedOnRetry()
```
robshakespeare, line 507: **praise:** Very nicely implemented, clean and easy to follow :+1:

robshakespeare (changes requested):
Fantastic set of changes.

Thank you for updating the tests to be behavioural too, in this case of verifying security logic it is the right approach 👍 

Please can you complete the requested changes, there's a few but they're only minor.  Thank you.

robshakespeare (approved):
Fantastic, thank you for making all of these changes and the attention to detail.

PR #667, opened 2026-06-09
feat(find): Update Find API to optionally authorise scopes directly from Auth Store rather than token scopes

jordanstarksbrowning-hippo:
## Summary

We need to update the Find API to optionally authorise scopes directly from the Auth Store rather than token scopes, so that we can support integrating with DfE’s Find and Use an API (FaUAPI).

FaUAPI supports multiple scopes, but they cannot differ between client/user/subscription, so cannot be used to authorise different roles or levels of access within an API.  A feature request has been submitted, but is unlikely to be implemented soon.

The goal of the ticket is to update the Find API so that the Auth Store can be used to authorise scopes/roles within the app, rather than token scopes.

## Changes

- Add new auth config value to use auth store for authorisation
- Implement logic to optionally use the Auth store to populate the scopes in the Auth Context, instead of the token's scopes
- Add new tests to cover the new methods

## Validation


Apps/Find/src/SUI.Find.FindApi/Middleware/AuthContextFactory.cs
```diff
@@ -1,12 +1,16 @@
...
+using SUI.Find.Infrastructure.Services;
 
 namespace SUI.Find.FindApi.Middleware;
 
 public class AuthContextFactory : IAuthContextFactory
 {
-    public AuthContext FromJwt(JwtSecurityToken jwt, AuthStore store)
+    public AuthContext FromJwt(
+        JwtSecurityToken jwt,
+        IAuthStoreService storeService,
```
robshakespeare: **nitpick:** Please can this be changed to be an injected service.  I think that should be possible now that the `AuthContextFactory` is a DI service.

Apps/Find/src/SUI.Find.Infrastructure/Services/MockAuthStoreService.cs
```diff
@@ -27,6 +28,42 @@ public Task<AuthStore> GetAuthStoreAsync()
         return Task.FromResult(_authStore.Value);
     }
 
+    public IReadOnlyList<string> GetScopesByClientId(string clientId)
+    {
+        var clients = GetClients();
+
+        var scopes = clients.FirstOrDefault(x => x.ClientId == clientId)?.AllowedScopes ?? [];
```
robshakespeare: todo: Please can the logic to get a client by ID be extracted in to a method, so that the 2 places in this class use that same method.  The method can be private.

Also, I think in the context of security, we should throw an exception if not found, rather than have graceful degradation.

i.e. something like:
```
private AuthClient GetClientById(string clientId) =>
    GetClients().FirstOrDefault(x => x.ClientId == clientId)
    ?? throw new InvalidOperationException($"Client {clientId} could not be found in auth store.");
```

And then the 2 places with the duplicated lookup logic can use that method instead.

Also, please could you update the `MockAuthStoreServiceTests` to add new tests to verify that the exception is thrown when expected.

Many thanks

Apps/Find/src/SUI.Find.Infrastructure/Services/SecureAuthStoreService.cs
robshakespeare, line 1: **suggestion:** I think its fine to just delete `SecureAuthStoreService` for now.

Apps/Find/tests/SUI.Find.Infrastructure.UnitTests/Services/MockAuthStoreServiceTests.cs
```diff
@@ -77,4 +86,36 @@ public async Task LoadStore_ShouldThrowInvalidOperationException_WhenJsonIsCorru
...
+        // Act
+        var result = _sut.GetScopesByClientId("LOCAL-AUTHORITY-01");
+
+        // Assert
+        Assert.NotNull(result);
+        Assert.Equivalent(ExpectedScopes, result, strict: true);
+    }
+
+    [Fact]
+    public async Task GetOrganisationIdForClientId_WithValidClientId_ShouldReturnScopes()
```
robshakespeare: **nitpick:** Test method name typo:

```suggestion
    public async Task GetOrganisationIdForClientId_WithValidClientId_ShouldReturnOrganisationId()
```

Apps/Find/tests/SUI.Find.FindApi.UnitTests/MiddlewareTests/AuthContextFactoryTests.cs
```diff
@@ -15,70 +17,60 @@ public void TestFromJwt_WhenClientIdAndSubEmpty_ShouldThrowException()
...
-            },
+            new("client_id", clientId),
+            new("scope", "file.read file.write"),
         };
+        var jwt = new JwtSecurityToken(claims: claims);
 
-        var store = new AuthStore { Clients = clients, DefaultTokenLifetimeMinutes = 60 };
+        var store = Substitute.For<IAuthStoreService>();
+        store.GetOrganisationIdForClientId("EDUCATION-01").Returns(organisationId);
+        store.GetScopesByClientId(Arg.Any<string>()).Throws<InvalidOperationException>(); // Should not use Auth Store for authorisation
```
robshakespeare: **praise:** Nice, good thinking for a solid test, same with the "incorrect.scope" bit below :+1:

robshakespeare (commented):
Thanks for making the updates.  Just a few requested changes please, all fairly minor.  Thank you

Apps/Find/src/SUI.Find.Infrastructure/Services/MockAuthStoreService.cs
```diff
@@ -27,6 +28,39 @@ public Task<AuthStore> GetAuthStoreAsync()
...
 
+    public IReadOnlyList<string> GetScopesByClientId(string clientId)
+    {
+        var client = GetClientById(clientId);
+        return client.AllowedScopes ?? [];
+    }
+
+    public string GetOrganisationIdForClientId(string clientId)
+    {
+        var clients = GetClientById(clientId);
```
robshakespeare: **nitpick:** Sorry to be super picky! The "s" has been left, please can this variable be renamed to be singular:

```suggestion
        var client = GetClientById(clientId);
```

Apps/Find/src/SUI.Find.Infrastructure/Services/MockAuthStoreService.cs
```diff
@@ -27,6 +28,39 @@ public Task<AuthStore> GetAuthStoreAsync()
...
+        return clients.OrganisationId;
+    }
+
+    private AuthClient GetClientById(string clientId)
```
robshakespeare, line 43: **praise:** Good thinking to combine `GetClients` with the new logic :+1:

Apps/Find/src/SUI.Find.Infrastructure/Services/MockAuthStoreService.cs
robshakespeare: Many thanks for making this update

Apps/Find/src/SUI.Find.Infrastructure/Services/SecureAuthStoreService.cs
robshakespeare: Thank you for deleting it

robshakespeare (changes requested):
Many thanks for making the updates.  Just one super small change please, then I think its good to go.  Thank you

Apps/Find/src/SUI.Find.Infrastructure/Services/MockAuthStoreService.cs
jordanstarksbrowning-hippo: Blast! Thanks for spotting that :D

robshakespeare (approved):
Super, thank you :+1:

Apps/Find/src/SUI.Find.FindApi/Middleware/AuthContextFactory.cs
robshakespeare: Thanks for updating

PR #668, opened 2026-06-09
feat(authEmulator): SUI-1779 - Sign tokens using RSA private key

udara-jay:
## Summary

Sign emulated tokens using RSA asymmetric keys, to avoid auth bypasses becoming part of the architecture.

## Changes

The SUI.AuthEmulator.Services.JwtTokenService service is updated to sign our emulated tokens using RSA asymmetric keys:
> The JwtTokenService randomly chooses an RSA key to sign with from the class-level service created in SUI-1776 (Add JWKS endpoint to Auth Emulator)
> When creating the JwtSecurityToken the SigningCredentials of the RSA key chosen above is used
> The JwtTokenService no longer uses the old SymmetricSecurityKey approach
> The AuthStore should no longer be used for configuration:
> - AuthSettings:Issuer retrieved from .NET config, the default should be: "https://sandbox.api.example.gov.uk/sui-find-a-record/auth"
> - AuthSettings:Audience retrieved from .NET config, the default should be: "sui-find-a-record-api"
> - AuthSettings:TokenLifetimeMinutes retrieved from .NET config, the default should be 60
> - The config values are loaded via a .NET Options Pattern object called AuthSettings

Also includes changing the e2e ephemeral workflow to use the .NET HTTPS development certificate, because the OIDC discovery requires HTTPS URLs, also it makes this workflow tidier.

## Validation

Successful Unit and E2E test runs


robshakespeare (approved):
Many thanks for making these changes :+1:

udara-jay:
> Many thanks for making these changes 👍

Thanks for your assistance on figuring out the certificate issue 😅

PR #670, opened 2026-06-12
refactor: SUI-1790 - drive AuthSettings via GitHub Environment Secrets

robshakespeare:
## Summary

This PR updates the AuthSettings to be driven via GitHub Environment Secrets, so that the values used in the infrastructure are hidden from the public repository, and so that the values can be different to enable integrating with real identity providers, for example FaUAPI.

## Changes

* Updated the GitHub Workflows to pass the AuthSettings from GitHub Environment Secrets to the Terraform.
* Also updated `Apps/AuthEmulator/src/SUI.AuthEmulator/appsettings.json` to include the default `Audience`.  This has no functional change, it just standardises the approach to match the other config values.
* Also includes minor fix in the e2e FunctionTestFixture: if the URL is rooted, only log the URL because the base address won't be used in that case.
* Also includes small update to main readme: added overview of each Development Environment.

## Validation

- IDE linting
- E2E test run verifications
- Terraform plans, to show the expected changes:
    - `Find` and `AuthEmulator` apps show that the only changes are the `AuthSettings__Issuer` and `AuthSettings__Audience` (because these values are now infrastructure specific, are hidden from the public code by coming from GitHub Env Vars)
    - `StubCustodians` and `UITestHarness` show no changes

### DIFF
--- a/.github/workflows/find-e2e-tests.yml
+++ b/.github/workflows/find-e2e-tests.yml
@@ -216,7 +216,9 @@ jobs:
     # Runner labels from RUNNER_LABELS (JSON array). Default: ubuntu-latest.
     runs-on: ${{ fromJSON(vars.RUNNER_LABELS || '["ubuntu-latest"]') }}
     if: needs.eval-workflow-setup.outputs.run_dev == 'true'
-    environment: ${{ inputs.environment || 'd01' }}
+    environment:
+      name: ${{ inputs.environment || 'd01' }}
+      deployment: false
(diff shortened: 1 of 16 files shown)


udara-jay (approved):
Thank you for doing this! Looks much cleaner and clearer 👍🏽

robshakespeare:
> Thank you for doing this! Looks much cleaner and clearer 👍🏽

Thank you :+1:

PR #671, opened 2026-06-15
refactor: sui-1780 - Remove Inbound Symmetric Key & verification logic

udara-jay:
## Summary

- Remove the unused SymmetricSecurityKey and related code, that was obsoleted to avoid auth bypasses becoming part of the architecture.

## Changes

- The SUI.StubCustodians.API.ScopeEnforcementMiddleware and SUI.StubCustodians.API.Controllers.AuthController classes are changed to use a copy of SUI.StubCustodians.Application.Models.AuthStore class but called OutboundAuthStore, and that copied class should keep all of the defined properties.
- The signingKey, issuer, audience and defaultTokenLifetimeMinutes properties are removed from Data/auth-clients-inbound.json
- The SigningKey, Issuer, Audience and DefaultTokenLifetimeMinutes properties are removed from the models:
> SUI.AuthEmulator.Models.AuthStore
> SUI.StubCustodians.Application.Models.AuthStore
- The “Auth store file is missing issuer, audience, or signingKey” validation logic is removed from MockAuthStoreService
- The SymmetricSecurityKey branch of code in JwtAuthMiddleware (that was used for token verification, but now isn’t used anymore) is totally removed, and any related now unused code is also removed.

## Validation

- Successful Unit tests and E2E tests

### DIFF
--- a/Apps/AuthEmulator/src/SUI.AuthEmulator/Services/MockAuthStoreService.cs
+++ b/Apps/AuthEmulator/src/SUI.AuthEmulator/Services/MockAuthStoreService.cs
@@ -25,17 +25,6 @@ public class MockAuthStoreService : IAuthStoreService
     {
         var store = _authStore.Value;
 
-        if (
-            string.IsNullOrWhiteSpace(store.Issuer)
-            || string.IsNullOrWhiteSpace(store.Audience)
-            || string.IsNullOrWhiteSpace(store.SigningKey)
-        )
-        {
-            throw new InvalidOperationException(
-                "Auth store file is missing issuer, audience, or signingKey."
-            );
-        }
-
         store.Clients ??= [];
 
         var client = store.Clients.FirstOrDefault(c =>
--- a/Apps/AuthEmulator/src/SUI.AuthEmulator/Services/MockAuthStoreService.cs
+++ b/Apps/AuthEmulator/src/SUI.AuthEmulator/Services/MockAuthStoreService.cs
@@ -15,12 +15,6 @@ public class MockAuthStoreService : IAuthStoreService
         _authStore = new Lazy<AuthStore>(LoadStore);
     }
 
-    public Task<AuthStore> GetAuthStoreAsync()
-    {
-        // Instantly returns the cached in-memory store as a completed Task
-        return Task.FromResult(_authStore.Value);
-    }
-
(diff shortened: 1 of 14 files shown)


Apps/Find/tests/SUI.Find.FindApi.UnitTests/MiddlewareTests/JwtAuthMiddlewareTests.cs
```diff
@@ -319,122 +292,9 @@ public async Task TestInvoke_WithInvalidAuthHeaders_ReturnsProblemResponse()
...
-            );
-
-            // Act
-            await sut.Invoke(context, Next);
-
-            // Assert
-            AssertAccessAllowed(context, authContext);
-        }
-    }
+    // REMOVED: SymmetricVerificationTests class entirely
```
robshakespeare: **suggestion:** Looking at all of this now, I think it would be good to keep the `SymmetricVerificationTests` subclass, but have just one test, that verifies given a symmetric token, we get an access denied response.

So the test would be something like:

```
// The arrange here would do the setup that includes GenerateSymmetricToken

// Act
await sut.Invoke(context, Next);

// Assert
AssertAccessDenied(context, "Token validation failed", "Signature validation failed");
```

Apps/AuthEmulator/src/SUI.AuthEmulator/Services/MockAuthStoreService.cs
robshakespeare, line 1: **todo:** Just checked through, there is one more thing that can be removed now.  Please can you also remove `GetAuthStoreAsync` in all cases:

* `SUI.AuthEmulator.Services.IAuthStoreService.GetAuthStoreAsync()`
* `SUI.AuthEmulator.Services.MockAuthStoreService.GetAuthStoreAsync()`
* `SUI.Find.Infrastructure.Services.IAuthStoreService.GetAuthStoreAsync()`
* `SUI.Find.Infrastructure.Services.MockAuthStoreService.GetAuthStoreAsync()`

Thanks

robshakespeare (commented):
Thank you for doing this.  Just a couple of small suggestions please.  Thank you

robshakespeare (approved):
Great, many thanks

PR #672, opened 2026-06-15
feat(find): SUI-1775 - Update Auth Store (auth-clients-inbound.json) to use Client ID values that do not match Organisation ID

jordanstarksbrowning-hippo:
## Summary
As part of evolving the Auth to be more production-ready we need to update Auth Store (auth-clients-inbound.json) to use Client ID values that do not match their respective Organisation ID, so that we can verify that the Client ID can be any value to enable integrating with 3rd party identity providers.

## Changes

- Add "CLIENT-ID_" as a prefix to existing client IDs in the auth json file
- Update tests to search for these values in the auth store
- Update tests to follow the new client ID pattern for consistency

## Validation

- Unit/integration/E2E tests run

### DIFF
--- a/Apps/Find/tests/SUI.Find.FindApi.UnitTests/MiddlewareTests/AuthContextFactoryTests.cs
+++ b/Apps/Find/tests/SUI.Find.FindApi.UnitTests/MiddlewareTests/AuthContextFactoryTests.cs
@@ -9,6 +9,8 @@ namespace SUI.Find.FindApi.UnitTests.MiddlewareTests;
 
 public class AuthContextFactoryTests
 {
+    private const string ClientId = "CLIENT-ID_EDUCATION-01";
+    private const string OrganisationId = "EDUCATION-01";
     private readonly AuthContextFactory _sut;
     private readonly IAuthStoreService _store;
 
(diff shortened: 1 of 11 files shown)


Apps/StubCustodians/src/SUI.StubCustodians.Application/Utilities/CustodianWorker.cs
```diff
@@ -159,7 +159,7 @@ CancellationToken ct
         var baseUrl = _config["StubCustodians:BaseUrl"]!;
 
         var manifest = await manifestService.GetManifestForOrganisation(
-            _authClient.ClientId,
+            _authClient.OrganisationId,
```
robshakespeare, line 162: **praise:** Excellent find, thank you for investigating this and tracking it down! ❤️

Apps/Find/tests/SUI.Find.FindApi.UnitTests/MiddlewareTests/AuthContextFactoryTests.cs
```diff
@@ -9,6 +9,8 @@ namespace SUI.Find.FindApi.UnitTests.MiddlewareTests;
 
 public class AuthContextFactoryTests
 {
+    private const string ClientId = "CLIENT-ID_EDUCATION-01";
+    private const string OrganisationId = "EDUCATION-01";
```
robshakespeare, line 13: **praise:** Nice, that's much clearer now :+1:

robshakespeare (approved):
Excellent, thank you

PR #673, opened 2026-06-15
feat(find): SUI-1791 - Support for azp (Authorized party) claim to be used as the Client ID

udara-jay:
## Summary

Support the azp (Authorized party) claim to be used as the Client ID, so that we can integrate with various real identity providers.
The azp claim is used by the DfE Find and Use an API (FaUAPI) service to represent each “Client” (known as Subscriptions in FaUAPI’s terminology).

## Changes

- The AuthContextFactory is updated to try and get the “Client ID” value from the azp claim.
- The order of precedence of the claims is (from highest to lowest precedence):
> client_id 
> azp 
> sub
- Unit test is included the verifies that if both the azp and sub claims are present, that the azp claim is the value that is used as the Client ID.

## Validation

Successful unit tests


robshakespeare (approved):
Super, thank you

PR #678, opened 2026-06-16
feat: SUI-1785 - configure Non-public Client IDs and Secrets

robshakespeare:
## Summary

This PR enables non-public Client IDs and Secrets in our sample/mock data.

While having client secrets in a public repo is fine for local development and ephemeral environments, deployed environments should use actual secret values (rather than pretend secret values that have been publicly published) so that unauthorised people cannot authenticate with our deployed environments.

This will also enable us to be able to integrate with real identity providers, like Find and Use an API.

The design and implementation has been careful not to reveal any sensitive information in the public code or public workflow logs.

## Important Caveat

This design and implementation is a pragmatic step to completing the Auth Evolution, especially now that the scope of the programme is changing.

The final, tidier, and more straight forward to operate solution would be to remove our mock Organisation Directory and design and implement real a Organisation Directory, probably using Azure App Configuration.  However, it is not feasible to do that at this point in the project.

## Changes

- Added ability to override the Client IDs and Secrets in the Auth Store data, driving these values from JSON maps ultimately from GitHub Environment Secrets.
- Update sample data client IDs to not use dashes/hyphens.  Unfortunately hyphens are not supported as environment variable names in Azure App Settings, so the most straight forward solution is to sanitise our client IDs at source.
- Updated readme and added script to help explain and ease setup.

Also includes:
- `CustodianWorker` update: switch some more logging across to where its more useful to also see the Organisation ID.

## Validation

- Terraform plans
- Unit and E2E test runs
- Review of logs, to ensure no sensitive information is exposed:

image

image

image

### DIFF
--- a/.github/workflows/find-e2e-tests.yml
+++ b/.github/workflows/find-e2e-tests.yml
@@ -124,6 +124,8 @@ jobs:
     runs-on: ${{ fromJSON(vars.RUNNER_LABELS || '["ubuntu-latest"]') }}
     if: needs.eval-workflow-setup.outputs.run_ephemeral == 'true'
     env:
+      AuthClientCredentials__CLIENT-ID_LOCAL-AUTHORITY-01__NewClientId: TestExampleClientId-01
+      AuthClientCredentials__CLIENT-ID_LOCAL-AUTHORITY-01__NewClientSecret: TestExampleSecret
       # Point OpenSSL to the .NET dev certs trust directory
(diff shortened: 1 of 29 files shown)


udara-jay (approved):
Looks fantastic! Thank you for this brilliant set of changes 👍🏽

robshakespeare:
> Looks fantastic! Thank you for this brilliant set of changes 👍🏽

Thanks Udara 😊

PR #680, opened 2026-06-18
fix(find): SUI-1804 - Fix Find's Terraform, so that `plan` returns "no changes" when there are really no actual changes

jordanstarksbrowning-hippo:
## Summary

Some tweaks to the terraform for Find so that fields are not regenerated unless they actually change, to reduce noise ad make genuine changes easier to detect.

## Changes

- Remove entries written explicitly in AppSettings when they are re-applied by terraform automatically (See Notes under https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_function_app#app_settings-1)
- Refactor metrics to be explicitly generated dynamically on plan, so that terraform expects the changes

## Validation

- Terraform plan on d01


jordanstarksbrowning-hippo:
Screenshot 2026-06-18 at 13 29 46

robshakespeare (approved):
Excellent, thanks very much for working all of this out.

I'm happy to get this merged, and then validate on the azure portal that the expected values are still present after you've run apply, like you suggested :+1:

PR #681, opened 2026-06-18
ops: SUI-1786 - Provision d03 environment to test evolved Auth integrating with FaUAPI

udara-jay:
## Summary

Provision d03 "sandbox" environment, so that we can verify the generic OAuth2/JWT auth model (Auth Evolution) by integrating the new deployed Sandbox environment with Find and Use an API (FaUAPI).

## Changes

- Any occurrence of the list of environment options in the GitHub workflows is updated to include a - d03 option.
- The d03 must not have the infrastructure for the Auth Emulator deployed (use_auth_emulator should be false in the d03 Terraform)
- Terraform environment variables are created:
> Terraform/environments/d03.tfvars is created as a copy of d01.tfvars with the following changes:
> - environment_id = "d03"
> - use_auth_emulator = false
- The d03.tfvars find_app_settings is updated to include AuthSettings__UseAuthStoreForAuthorisation = True
- The GitHub workflows have been ran to provision the d03 environment
> plan first, then apply once it looks correct, and the blocked tickets are complete.
> d03 is the target environment, and the workflows are applied in this order: 
> 1. Terraform Core Infrastructure
> 2. Terraform Find
> 3. Terraform Custodians
> 4. Terraform UI Harness
> 5. Build, Test, Deploy: "Find" Solution
> 6. Build, Test, Deploy: "Custodians" Solution
> 7. Build, Test, Deploy: "UI Test Harness" Solution

## Validation

Terraform Plan Summary content


robshakespeare (commented):
Looks spot on, thanks :+1:

robshakespeare (approved):
Solid set of changes, thank you

PR #683, opened 2026-06-19
feat(find): sui-1841 - Add Auth Client Enabled check

udara-jay:
## Summary

Check the authenticated client’s enabled flag when authorising access in the Find API; so that access can be disabled internally within the system, regardless of and separate to any 3rd party / external Identity Providers.

## Changes

- The AuthClient.Enabled flagged is checked within the Find API’s authentication and authorisation process, and if the authenticated Auth Client is disabled then the API should deny access and respond with 401 Unauthorized.
- The check is only performed in `JwtAuthMiddleware`, a single automatic place, so that we do not have to remember to include this check, and so that we are certain it is always being checked.

## Validation

Successful unit tests for both enabled clients being authorised and disabled clients being denied access.

### DIFF
--- a/Apps/Find/src/SUI.Find.FindApi/Models/Auth/AuthResult.cs
+++ b/Apps/Find/src/SUI.Find.FindApi/Models/Auth/AuthResult.cs
@@ -1,0 +1,39 @@
+using System.Diagnostics.CodeAnalysis;
+
+namespace SUI.Find.FindApi.Models.Auth;
+
+public class AuthResult
+{
+    [MemberNotNullWhen(true, nameof(Context))]
+    [MemberNotNullWhen(false, nameof(FailureType), nameof(ErrorMessage))]
+    public bool IsSuccess { get; }
+    public AuthContext? Context { get; }
+    public AuthFailureReason? FailureType { get; }
+    public string? ErrorMessage { get; }
+
+    private AuthResult(
+        bool isSuccess,
+        AuthContext? context,
+        AuthFailureReason? failureType,
+        string? errorMessage
+    )
+    {
+        IsSuccess = isSuccess;
+        Context = context;
+        FailureType = failureType;
+        ErrorMessage = errorMessage;
+    }
+
+    public static AuthResult Success(AuthContext context) => new(true, context, null, null);
+
+    public static AuthResult Failure(AuthFailureReason reason, string errorMessage) =>
+        new(false, null, reason, errorMessage);
+}
+
+public enum AuthFailureReason
+{
+    InvalidTokenClaims,
+    ClientNotFound,
+    ClientDisabled,
+    MissingOrganisationId,
+}
--- a/Apps/Find/src/SUI.Find.FindApi/Models/AuthContext.cs
+++ b/Apps/Find/src/SUI.Find.FindApi/Models/AuthContext.cs
@@ -3,6 +3,5 @@ namespace SUI.Find.FindApi.Models;
 public sealed record AuthContext(
     string ClientId,
     string OrganisationId,
-    IReadOnlyList<string> Scopes,
-    bool IsEnabled
+    IReadOnlyList<string> Scopes
 );
--- a/Apps/Find/src/SUI.Find.FindApi/Models/AuthContext.cs
+++ b/Apps/Find/src/SUI.Find.FindApi/Models/AuthContext.cs
@@ -3,5 +3,6 @@ namespace SUI.Find.FindApi.Models;
 public sealed record AuthContext(
     string ClientId,
     string OrganisationId,
-    IReadOnlyList<string> Scopes
+    IReadOnlyList<string> Scopes,
+    bool IsEnabled
 );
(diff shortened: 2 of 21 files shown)


Apps/Find/src/SUI.Find.FindApi/Models/AuthContext.cs
```diff
@@ -3,5 +3,6 @@ namespace SUI.Find.FindApi.Models;
 public sealed record AuthContext(
     string ClientId,
     string OrganisationId,
-    IReadOnlyList<string> Scopes
+    IReadOnlyList<string> Scopes,
+    bool IsEnabled
```
robshakespeare: **nitpick:** I'm not overly keen on allowing the `AuthContext` to be valid or invalid.  Please can I suggest we refactor the `AuthContextFactory` so it can return 3 different states.

Maybe a tuple or OneOf (or maybe even the call to `GetClientById` could be done directly from the `JwtAuthMiddleware`, and therefore the first 2 states are handled directly in `JwtAuthMiddleware`?  What ever you think will be best, but the goal is to not have the AuthContext.IsEnabled prop).

The 3 states that I think will be helpful to differentiate:

1. No matching client in Auth Store
2. Client is disabled
3. AuthContext

Thank you

Apps/Find/src/SUI.Find.Infrastructure/Services/MockAuthStoreService.cs
```diff
@@ -36,6 +37,12 @@ public string GetOrganisationIdForClientId(string clientId)
         return client.OrganisationId;
     }
 
+    public bool IsClientEnabled(string clientId)
+    {
+        var client = GetClientById(clientId);
```
robshakespeare: **suggestion:** Looking at all of this now, please can you refactor this and the existing code.  My suggestion is to remove the `GetScopesByClientId`, `GetOrganisationIdForClientId` and `IsClientEnabled` methods, and instead expose the `GetClientById` method.  That way the `GetClientById` logic will only be done once, which will be a small performance gain, and then the properties `Enabled`, `OrganisationId` and `Scopes` can be used directly in the `AuthContextFactory`.

robshakespeare (commented):
PR looks good, but now I look at the whole thing including the existing code, please can I make some suggestions for a refactor.  Thank you 🙂

Apps/Find/src/SUI.Find.FindApi/Models/Auth/AuthResult.cs
```diff
@@ -0,0 +1,39 @@
+using System.Diagnostics.CodeAnalysis;
+
+namespace SUI.Find.FindApi.Models.Auth;
+
+public class AuthResult
+{
+    [MemberNotNullWhen(true, nameof(Context))]
+    [MemberNotNullWhen(false, nameof(FailureType), nameof(ErrorMessage))]
```
robshakespeare, line 8: **praise:** Nice, good thinking with the `AuthResult` approach and using `MemberNotNullWhen` :+1:

Apps/Find/tests/SUI.Find.FindApi.UnitTests/MiddlewareTests/AuthContextFactoryTests.cs
```diff
@@ -118,16 +180,17 @@ public void TestFromJwt_WithValidInputs_UsingTokenScopes_ReturnsAuthContext()
         };
         var jwt = new JwtSecurityToken(claims: claims);
 
-        _store.GetOrganisationIdForClientId(ClientId).Returns(OrganisationId);
-        _store.GetScopesByClientId(Arg.Any<string>()).Throws<InvalidOperationException>(); // Should not use Auth Store for authorisation
+        var validClient = new AuthClient { Enabled = true, OrganisationId = OrganisationId };
```
robshakespeare: **nitpick:** Please could you change this to also initialise the client's AllowedScopes property to a different value, e.g.

```suggestion
        var validClient = new AuthClient
        {
            Enabled = true,
            OrganisationId = OrganisationId,
            AllowedScopes = ["incorrect.scope"], // Should not read from here, should read from token
        };
```

Just to be more explicit in the tests.

Apps/Find/tests/SUI.Find.FindApi.UnitTests/MiddlewareTests/JwtAuthMiddlewareTests.cs
```diff
@@ -17,6 +17,7 @@
 using SUI.Find.FindApi.Configurations;
 using SUI.Find.FindApi.Middleware;
 using SUI.Find.FindApi.Models;
+using SUI.Find.FindApi.Models.Auth; // <-- Added this using directive
```
robshakespeare: **nitpick:** Was the comment accidentally left in?

```suggestion
using SUI.Find.FindApi.Models.Auth;
```

robshakespeare (dismissed):
Nice, many thanks for making the updates.

Just a couple of minor points, but not blocking.

robshakespeare (approved):
Looks good, thank you

PR #687, opened 2026-06-23
ops: SUI-1842 - Ensure the App Insights connection string cannot be exposed in public logs

udara-jay:
## Summary

Mark the App Insights connection string as sensitive in the Terraform, so that the instrumentation key is redacted from the pubic GitHub workflow logs.

## Changes

- The Terraform is updated to wrap the App Insights connection string in a call to the sensitive function.

- The call to the sensitive function should be done in a central place, so that we do not need to remember to have to do it anywhere the connection string is used, and so that we ensure it is always marked as sensitive. 

## Validation

Plan and apply Terraform summaries.

### DIFF
--- a/terraform/auth-emulator/main.tf
+++ b/terraform/auth-emulator/main.tf
@@ -43,9 +43,11 @@ module "web_app" {
   service_offering = var.service_offering
 
   dotnet_version = var.webapp_dotnet_version
+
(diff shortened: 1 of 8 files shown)


robshakespeare (approved):
Thank you for sorting this :+1:

PR #688, opened 2026-06-23
fix: SUI-1829 - fixed correlation of logging and tracing across disconnected processes

robshakespeare:
## Summary

The main change in this PR fixes correlation of logging and tracing across disconnected processes.  The missing piece to the puzzle was that the "sub-activities" need to be added as a trace source to the Open Telemetry package.  The "sub-activities" are necessary to be able to specify the Trace ID to correlate the disconnected processes.

This PR also includes lots of minor related fixes and improvements that were discovered along the way to fixing the main issue.

## Changes

- Fixed the main problem (correlation of logging and tracing across disconnected processes) by instructing the OpenTelemetry package to listen to the activity source of our sub-activities.
  * Additional fix was to specify the activity kind (Internal vs Consumer), so that when the trace is viewed in App Insights it makes more sense.
  * Third fix was to change the "Started activity" log message to debug level, so that it doesn't create noise in the logging now that the correlation is all working as intended.

Also includes:
- Display Trace ID and trace info/link of the Find process in the debug info panel in the UI Test Harness.
    * Related to this: there were various issues caused by attempting to generate a working App Insights link in the E2E Tests, so instead it has been changed to log out the App Insights query so its easy to just copy it and use it directly.
- Include the log scope metadata as tags in the sub-activities, so that they are included in the Trace Visualizer in App Insights.
- Tidied up the run profiles so that it is easy to launch typical setups.  We are not able to test the system locally without Find+Stubs+AuthEmu running, and the only real option is the UI.  The old profiles didn't make sense anymore.  These changes make it more straight forward to test locally.
- Changed level of existing noisy log statements from Info down to Debug, so that they excluded by default.
- Improved various log messages to avoid confusion when looking an entire trace, or a large amount of logs.
- Include `TraceParent` in the HTTP auditing, all the way down to the table storage, to enrich audit logs with additional helpful info.

## Validation

- Unit and E2E tests
- Manually run locally, connected to App Insights, to ensure that the correlation of logging and tracing across disconnected processes works as expected


udara-jay (approved):
Amazing set of changes! Thank you for making everything much more cohesive and clear!

PR #690, opened 2026-06-24
Ensure d03 works with FAUAPI

jordanstarksbrowning-hippo:
## Summary

This PR contains the final changes required to make the system work behind a gateway, like d03 using FaUAPI.

## Changes

- Added a E2E tests workflow step to run the E2E tests when the Find API is behind a gateway.
- Updated the relevant components and areas to be able to optionally use the Find API Gateway URL and Auth Scopes, and updated related places where the `api/` route prefix should be part of the base URL, because the prefix is not included when running through a gateway.
- Added trace diagnostics to the workers in the Stub Custodians, to help debug issues in the stubs.

## Validation

- Lots of E2E tests runs! Against d03 and ephemeral.


robshakespeare (approved):
Thank you for working on this and preserving with the debugging and fixes.

Great to see now have a green e2e test run against `d03`!  https://github.com/DFE-Digital/single-unique-identifier/actions/runs/28225019128

PR #691, opened 2026-06-24
feat: SUI-1848 - resolve Authentication issues in E2E tests when a gateway (e.g. FaUAPI) is being used

robshakespeare:
## Summary

This PR resolves the issues in the E2E Tests when they run against an environment that has a gateway, like d03 with FaUAPI as the gateway.

The issues fixed in this PR are:
* Going via gateway needs different scopes
* Going via gateway needs Auth token for Find API's health check

## Changes

- Ability to override the auth scope in the E2E tests, to support running the tests through FaUAPI or other gateways.
- Refactored E2E tests to make it possible to reuse the Access Token logic.
- Update the Find API health check to attach an access token to the request.  Although this isn't needed for our lower dev environments, it is needed when running against environments that use FaUAPI or other gateways.

Also includes:
- Removed the `SearchTestsBase.TestScopes` field, because it isn’t needed.
- Refactored the E2E health check logic to use the XUnit test output helper in all cases, rather than diagnostic messages.  The problem with diagnostic messages is they come out as warnings which makes things look far more broken than they really are.

## Validation

- Manual runs of E2E tests locally
- PR validation
- Manual run of E2E workflow against d01, to ensure no breaking changes in the E2E test logic when running against deployed environments.


jordanstarksbrowning-hippo (approved):
Thanks for this Rob! Looks good

PR #693, opened 2026-06-25
ops: SUI-1806 - Add toggle for UI Test Harness infrastructure deployment

udara-jay:
## Summary

Add the ability to toggle the UI Test Harness infrastructure off per environment in the Terraform; so that we can deploy later stage environments where they Test Harness will not be needed, as we move towards support for Alpha.

## Changes

- Add a new `use_ui_test_harness` boolean in the `ui_test_harness` Terraform module, that toggles whether the UI Test Harness infrastructure gets deployed.
- `use_ui_test_harness` defaults to false.
- `use_ui_test_harness` is set to true in `d01.tfvars` and `d02.tfvars` and `d03.tfvars` (if `d03.tfvars` exists at the point this ticket is done).
- When `use_ui_test_harness` is false then none of the resources specified in `ui_test_harness/main.tf` gets deployed.
> In this case, because there are so many resources, the existing contents of `ui_test_harness/main.tf `are refactored in to a new module within `ui_test_harness` so that the toggle is only applied to a single top-level module call in `ui_test_harness/main.tf`

## Validation

Successful terraform plan summaries


terraform/ui_test_harness/outputs.tf
```diff
@@ -1,19 +1,19 @@
 output "service_state_key" {
-  value       = local.service_state_key
+  value       = format("%s/ui_test_harness.tfstate", var.environment_id)
```
robshakespeare: **nitpick:** Would it be possible to make `service_state_key` an output of the new app module, and use that here, rather than have a duplicated value?

I realise this is only a minor issue, but I feel like we should sill avoid duplication where possible.

terraform/ui_test_harness/outputs.tf
```diff
@@ -1,19 +1,19 @@
...
   description = "State key for the UI Test Harness Service"
 }
 
 output "web_app_name" {
-  value       = module.web_app.name
+  value       = try(module.app[0].web_app_name, null)
```
robshakespeare: **nitpick:** I think that using the TF `one` function here is more of a standard approach in this case:


```suggestion
  value       = one(module.app[*].web_app_name)
```

robshakespeare (commented):
Thanks for doing these changes so far.

Bit of a pain having to duplicate and pass all of those variables isn't it.  But I still think this is the right approach, because trying to toggle each resource individually would get even messier.

Have you tried a TF plan yet?  I think we might need to use a `moved` block, there's an example of this in the auth-emulator.  It might not need to be done in this case, I'm not sure.  The key thing is that we don't want the resources to be deleted and recreated, we want TF to understand they are actually the same thing.  

I hope that makes sense?  I'm happy to chat more about it if needed.  Thanks 🙂

robshakespeare (approved):
Ace, thank you for doing this refactor.

PR #695, opened 2026-06-26
chore: SUI-1856 - Improve FindApi BaseUrl config flexibility

robshakespeare:
## Summary

To make it easier to configure the system in the future -and avoid time consuming, tricky to debug errors- the system has been updated to support the Find API Base URL config value with or without a trailing slash.

## Changes

- Updated the places where the Find API base URL is used (Stubs, UI Test Harness & E2E Tests) to support the URL with or without the trailing slash.
- Also fixed the API prefix change for when the UI Test Harness locally.

## Validation

- Manual E2E tests runs and manual run of Test Harness

### DIFF
--- a/Apps/Find/tests/SUI.Find.E2ETests/Config.cs
+++ b/Apps/Find/tests/SUI.Find.E2ETests/Config.cs
@@ -2,9 +2,17 @@
 
 public record Config
 {
-    public string BaseUrl { get; init; } = "http://localhost:7182/api/";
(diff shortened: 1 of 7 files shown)


jordanstarksbrowning-hippo (approved):
Thanks Rob, this is much more robust now.

robshakespeare:
> Thanks Rob, this is much more robust now.

Thank you

PR #696, opened 2026-06-26
ops: SUI-806 - Improve ability to debug the E2E ephemeral tests

udara-jay:
## Summary

Configure the apps that run in the E2E ephemeral tests to use App Insights for their logging, so we can see what each app is doing and help with debugging and diagnosing the cause of failed tests.

## Changes

- The e2e-test-ephemeral job is updated to:
> Set the `APPLICATIONINSIGHTS_CONNECTION_STRING` environment variable to the value of the `app_insights_connection_string` from the Core Terraform state.
> Set the `OTEL_RESOURCE_ATTRIBUTES` environment variable to the value of `deployment.environment=ephemeral,service.namespace=MAIS Pilot 3,cloud.region=na`
- When the E2E ephemeral tests run, the logs from the apps can be seen in App Insights

## Validation

- Successful application insights logs from e2e ephemeral runs


udara-jay:
Wrong branch and PR

PR #697, opened 2026-06-26
ops: SUI-1844 - Improve ability to debug the E2E ephemeral tests

udara-jay:
## Summary

Configure the apps that run in the E2E ephemeral tests to use App Insights for their logging, so we can see what each app is doing and help with debugging and diagnosing the cause of failed tests.

## Changes

- The e2e-test-ephemeral job is updated to:
> Set the `APPLICATIONINSIGHTS_CONNECTION_STRING` environment variable to the value of the `app_insights_connection_string` from the Core Terraform state.
> Set the `OTEL_RESOURCE_ATTRIBUTES` environment variable to the value of `deployment.environment=ephemeral,service.namespace=MAIS Pilot 3,cloud.region=na`
- When the E2E ephemeral tests run, the logs from the apps can be seen in App Insights

## Validation

- Successful E2E ephemeral test logs in application insights

### DIFF
--- a/.github/workflows/find-e2e-tests.yml
+++ b/.github/workflows/find-e2e-tests.yml
@@ -20,7 +20,6 @@ permissions:
   contents: read
   actions: read
   checks: write
-  id-token: write
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


.github/workflows/find-e2e-tests.yml
```diff
@@ -125,6 +125,8 @@ jobs:
     runs-on: ${{ fromJSON(vars.RUNNER_LABELS || '["ubuntu-latest"]') }}
     if: needs.eval-workflow-setup.outputs.run_ephemeral == 'true'
     env:
+      APPLICATIONINSIGHTS_CONNECTION_STRING: ${{ secrets.EPHEMERAL_E2E_APP_CONNECTION_STRING }}
```
robshakespeare: **todo:** Sorry, this came from what I said before I corrected it, but please can you update the secret to be called `EPHEMERAL_E2E_APP_INSIGHTS_CONNECTION_STRING`


```suggestion
      APPLICATIONINSIGHTS_CONNECTION_STRING: ${{ secrets.EPHEMERAL_E2E_APP_INSIGHTS_CONNECTION_STRING }}
```

(obviously it'll need to be updated in the GitHub settings as well as the code here)

robshakespeare (changes requested):
Nice!

Please could you make the requested change.

Also, I have an idea about improving the bit in the E2E tests logging where we log out the observability info.  I'll chat to you about that, thanks.

robshakespeare (approved):
Super, thanks for doing this.  This will make a big difference when the ephemeral tests aren't working :+1:

udara-jay:
> Super, thanks for doing this. This will make a big difference when the ephemeral tests aren't working 👍

Definitely! 💯

PR #701, opened 2026-06-30
feat: SUI-1855 - enable UI Test Harness to work when Find API is behind a gateway

robshakespeare:
## Summary

The changes in this PR enable the UI Test Harness to work when the Find API is behind a gateway, and includes related tidying.

## Changes

* UI Test Harness changes:
    * Added optional Gateway Auth Scope to the UI Test Harness, to use when calling Find API.
    * Removed `SUI.UIHarness.Web.Services.FindService.Scopes` because its not necessary.
    * Standardised config approach to use constants, to make the code more cohesive and future proofed.  Decided not to use options pattern in this case because the additional overhead isn't justified in a test harness.

Also includes:
- Noticed and updated the Find Match API key to be rotated in d03 too.

## Validation

- Manually triggered deployments of Terraform and App to d02 and d03, and manually tested to verify everything works as expected, whether there is a gateway or not.
- E2E test run locally and via GitHub ephemeral workflow.

### DIFF
--- a/.github/rotate-keyvault-secret.manifest.json
+++ b/.github/rotate-keyvault-secret.manifest.json
@@ -42,7 +42,8 @@
         "enabled": true,
         "environments": [
           "d01",
-          "d02"
+          "d02",
+          "d03"
         ]
       }
     }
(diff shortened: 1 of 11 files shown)


udara-jay (approved):
Thank you for these changes! Looks great 👍🏽

robshakespeare:
> Thank you for these changes! Looks great 👍🏽

Thank you 🙂

PR #702, opened 2026-07-01
feat(E2E): SUI-1850 - Restrict retrying on E2E health check and access token steps to avoid unnecessary retries

jordanstarksbrowning-hippo:
## Summary

The retry policies on some areas of the E2E tests are too broad, so we are waiting through multiple retries even when non-recoverable errors occur.

This PR restricts the health check and access token endpoints to limit their retries to a subset of http request exceptions.

## Changes

- Don't retry health check endpoint if a non-http exception is thrown, or if specific http requests are thrown
- Don't retry access token endpoint if a non-http exception is thrown, or if specific http requests are thrown

## Validation

- Running E2E tests


udara-jay (approved):
Looks better now. Thanks for incorporating the whitelist 👍🏽

PR #703, opened 2026-07-02
feat(E2E): Add caching to access token retrieval for the E2E tests

jordanstarksbrowning-hippo:
## Summary

Add in-memory caching for the access token for the end to end tests, to minimise the number of calls needed to the auth endpoint.

## Changes

- Add memory cache to Function Test Fixture, to be used by the Access Token Provider
- Apply caching to the access tokens

## Validation

- E2E Tests - local, ephemeral, and in d01

### DIFF
--- a/Apps/Find/tests/SUI.Find.E2ETests/AccessTokenProvider.cs
+++ b/Apps/Find/tests/SUI.Find.E2ETests/AccessTokenProvider.cs
@@ -1,12 +1,15 @@
 ﻿using System.Net;
 using System.Net.Http.Headers;
 using System.Net.Http.Json;
+using System.Security.Cryptography;
+using System.Text;
 using System.Text.Json;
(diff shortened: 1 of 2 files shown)


jordanstarksbrowning-hippo:
The log line "Access token received ok" is called in the method that calls the Auth endpoint for the access token.
The log line "Access token cache used" is called when the new cache is accessed.

Ephemeral run, before changes:
Screenshot 2026-07-03 at 10 49 05

Ephemeral run, after changes:
Screenshot 2026-07-03 at 10 49 32
Screenshot 2026-07-03 at 10 51 18

Confirming that the number of calls to the Auth endpoint has reduced, and that the cache is being accessed heavily (100 is the maximum number that this search function can display).

udara-jay (approved):
Looks good 👍🏽

PR #704, opened 2026-07-03
test: SUI-1845 - E2E test coverage of authentication & authorisation

udara-jay:
## Summary

Expand the E2E tests to cover the authentication & authorisation behaviour, so that we are certain that the security checks are behaving as necessary.

## Changes

- E2E tests are written to cover the scenarios described below.
- The new E2E tests **only call** the ‘Get An Identifier’ endpoint (`MatchFunction`).
- The new E2E tests do **not** follow the V1 / V2 subclass pattern.
- The new E2E tests must **only use** the `CLIENT_ID_LOCAL_AUTHORITY_01` and `CLIENT_ID_INVALID_SCOPE` organisations.

**Scenarios**

> Valid token, should Allow access
> Valid token, but missing required scope, should Deny access
> Invalid signature due to modified token payload, should Deny access
> Invalid signature due to modified token header, should Deny access
> Invalid signature due to signature removal, should Deny access
> Invalid due to use of token before valid time range (i.e. not-before), should Deny access (* implementation-note)
> Invalid due to use of token after valid time range (i.e. expiration), should Deny access (* implementation-note)
> Invalid signature due to non-genuine private key, should Deny access (* implementation-note)
> Invalid source (i.e. issuer), should Deny access (* implementation-note)
> Invalid target (i.e. audience), should Deny access (* implementation-note)

**Implementation Note**

- The scenarios tagged with (* implementation-note) will require an update to the Auth Emulator’s token endpoint so that it can issue non-valid tokens.  The suggestion here is have a mode header sent in the request to the token endpoint, that can apply special logic during the token creation:

> ```
> mode=not-yet-active
> mode=expired
> mode=spoof-private-key
> mode=spoof-issuer
> mode=spoof-audience
> ```

- The tests for the scenarios that use the mode header should be tagged with `[Trait("RequiresAuthEmulator", "true")]`so that we can exclude them when we run the E2E tests against environments where the Identity Provider is not the Auth Emulator.

- The new Run E2E Tests (against dev environment via gateway) step in the E2E tests workflow should use the filter: `--filter "Category=E2E&Suite=Standard&RequiresAuthEmulator!=true" \`

## Validation

Successful E2E test runs

### DIFF
--- a/.github/workflows/find-e2e-tests.yml
+++ b/.github/workflows/find-e2e-tests.yml
@@ -272,7 +272,7 @@ jobs:
           E2E__AuthClientSecretsJsonMap: ${{ secrets.AUTH_CLIENT_SECRETS_JSON_MAP }}
         run: |
           dotnet test Apps/Find/tests/SUI.Find.E2ETests/SUI.Find.E2ETests.csproj \
-            --filter "Category=E2E&Suite=Standard" \
+            --filter "Category=E2E&Suite=Standard&RequiresAuthEmulator!=true" \
             --logger "console;verbosity=detailed" \
             --logger "trx;LogFileName=e2e-dev.trx" \
             --results-directory TestResults \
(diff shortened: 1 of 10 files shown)


jordanstarksbrowning-hippo (approved):
These tests are a great addition, and the new modes parameter for the Auth Emulator is a great idea. Nice one!

PR #708, opened 2026-07-07
Update dependabot configuration with cooldown settings

jtaylor100:
Reduces supply chain risks to the repository, by waiting some time before applying new versions, in case a supply chain attack occurs and remedial steps are taken.

Kept minimum value at 7, since dependencies are only checked weekly.

Lack of cooldowns raised by Hippo security reporting.


jordanstarksbrowning-hippo (approved):
Good addition, thanks Josh

PR #709, opened 2026-07-07
ops: Pass the API key in workflow to ensure failing E2E tests on d01

udara-jay:
## Summary

The new MatchAuth tests failed to obtain the FindApiKey when running against the d01 dev environment. Therefore additional steps were taken to access the key vault in the workflow and assign it where necessary for the tests.

## Changes

- Updated find-e2e-tests workflow to login and extract the required api key via the azure key vault.

## Validation

- Successful E2E tests against ephemeral as well as d01 (https://github.com/DFE-Digital/single-unique-identifier/actions/runs/28873785861).

### DIFF
--- a/.github/workflows/find-e2e-tests.yml
+++ b/.github/workflows/find-e2e-tests.yml
@@ -20,6 +20,7 @@ permissions:
   contents: read
   actions: read
   checks: write
+  id-token: write
 
 on:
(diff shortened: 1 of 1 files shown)


jordanstarksbrowning-hippo (approved):
Nice find, thanks for addressing this

PR #714, opened 2026-07-14
chore: SUI-1860 - Remove obsolete application code and tests

jordanstarksbrowning-hippo:
## Summary

We are deleting code from projects that will no longer be used after the most recent pivot. We are erring on the side of deleting more thoroughly, as we can always restore old projects using the git history.

## Changes

- Remove Transfer, Single View, Find, and Stub Custodians from codebase
- Remove github workflows related to these solutions

## Validation

- N/A


udara-jay (approved):
Thanks for doing this. Looks good to me 🤞🏽

PR #718, opened 2026-07-17
ops(getAnIdentifier): SUI-1914 - Add new workflows for Get An Identifier

udara-jay:
## Summary

Add workflows and IaC to the GetAnIdentifier solution so that coverage and unit test checks can be run via the pipeline and future deployments can be assisted with.

## Changes

- Add build, test and deploy Get An Identifier workflow
- Add test coverage check script
- Add test run settings
- Update tests
- Add new terraform workflows and code

## Validation

- Successful pipeline runs


jordanstarksbrowning-hippo (approved):
Looks good to me!

PR #720, opened 2026-07-23
feat: SUI-1874 - Add GetAnIdentifier Functionality added back from old MatchFunction code

jordanstarksbrowning-hippo:
## Summary

Our new Get An Identifier Function is based on the old Match Function in Find, and the SUI-Matcher in the pilots. This PR reinstates the existing code necessary for GetAnIdentifier to function as the previous implementation. Tweaks to the behaviour to reflect the new requirements will be added in a future PR.

## Changes

- Source code added from v0.1.2-discovery tagged version of the repo
- New folder structure has been added
- Names have been changed to reflect new project and different function name
- Unit tests added for imported classes

## Validation

- Unit tests run and passing


udara-jay (approved):
I have had a look at this and it seems alright. It is a massive PR, so I assume we will find if there are any issues while we add in the next set of components.

Thank you for getting this in 👍🏽 

P.S - Just realised that the rename of the previous project had not been fully completed and breaks workflows. Will fix that in a separate PR 😅

PR #721, opened 2026-07-24
fix: SUI-1938 - clean up post project rename leftovers and update workflow files

udara-jay:
## Summary

The reincorporation of code into GetAnIdentifier required some further older references to be renamed.

## Changes

- Updated broken workflows
- Updated broken launch settings
- Updated solution files

## Validation

- Successful build and test pipeline runs


jordanstarksbrowning-hippo (approved):
Thanks for finding these Udara, looks good to me

PR #722, opened 2026-07-28
feat(getAnIdentifier): SUI-1873 - Implement the Get an Identifier adapter to proxy matching requests to PDS

udara-jay:
## Summary

The purpose of this ticket is to implement the request to go to PDS directly. Previously in Pilot 1 and Alpha we used an algorithm to enhance matching, this is no longer needed now that PDS has MOAB. MOAB runs behind the PDS search so there is no change on our end.

## Changes

- Add new GetAnIdentifierService
- Clean up redundant namespaces
- Fix and update tests according to new scope

## Validation

- Successful passes in build workflows


jordanstarksbrowning-hippo (approved):
Looks good to me, thanks Udara.

PR #726, opened 2026-08-06
docs: SUI-1896 - Initial living document for automating FaUAPI

Keithin8a:
## Summary

Added a living document to discuss what we can automate using FaUAPI including questions we need to ask the FaUAPI team to determine the best approach

## Changes

- Added an initial version of the document

## Validation

- Docs-only change; no code/runtime validation required

### DIFF
--- a/Docs/Design/FaUAPI/AutomatedFaUAPI.md
+++ b/Docs/Design/FaUAPI/AutomatedFaUAPI.md
@@ -1,0 +1,132 @@
+# Automation Capabilities of Find and Use API
+
+**Date:** `2026-08-03`  
+**Owner:** SUI Service Team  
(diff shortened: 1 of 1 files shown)


IgiCodes (approved):
Happy to approve this for now ✅ 

Given it's a living document and should reflect the current state of things would be good to maybe mention that the function app doesn't currently expose the `/api/swagger.json` endpoint. We’ll need to add and configure the necessary OpenAPI extension before FaUAPI can import from that endpoint. If we're going to sort this imminently then happy to just merge this as-is.

It might also be worth softening the OAuth section from "there should be no change" to something we’ll confirm during the FaUAPI spike. I believe we still need to verify the issuer/audience configuration and whether this will use FaUAPI's application-restricted, machine-to-machine flow.

